PR #17, opened 2015-06-23
Added method matching support.

NOX73:


wheresrhys:
Sorry, I only just saw this *after* doing a big rewrite. But good idea, and I'd be happy to merge if you're able to resolve the new conflicts with master. The main difference in the code is the move to es6 (or, rather, the subset of it supported in nodejs4)

NOX73:
@wheresrhys I've rebased it

wheresrhys:
Excellent thanks

PR #19, opened 2015-09-11
Allow repeated calls to reset and restore

Lrnk:
Hi Rhys, I've done this hacky little fix to allow restore and reset to fail silently when they're not valid to call (called multiple times between new mocking behaviour specified?). I don't fully understand how it works, so you might be able to see a more elegant way to make this work.

Doing this allows me to put fetchMock.restore in the afterEach function instead of in the then and catch blocks of each test.

### DIFF
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -281,7 +281,9 @@ FetchMock.prototype.restore = function () {
 	this.isMocking = false;
 	this.reset();
 	if (this.usesGlobalFetch) {
-		theGlobal.fetch.restore();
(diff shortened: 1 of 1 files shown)


Lrnk:
Actually this is much less valuable than I thought. The problem comes in when restore/reset is called when it's not mocking, so Arjun pointed me more in this direction and now it's fine:

    afterEach(function() {
        if(fetchMock.isMocking){
            fetchMock.restore();
        }
    });

wheresrhys:
Just seen this, and have just done a big refactor so difficult to merge. cheers for the PR though

PR #43, opened 2015-12-09
check if matcher's first arg is Request

vbauerster:
A fetch could be called with Request param, instead of plain string url.
This code will fail to match
```js
fetchMock.mock('http://domain1', 200);
```
if fetch called like this:
```js
fetch(new Request('http://domain1'));
```
I've added the fix, which will match such calls.

### DIFF
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -59,7 +59,7 @@ function mockResponse (url, config) {
 
 function toRequest(url, options) {
 	var request;
-	if (Request.prototype.isPrototypeOf(url) && !options) {
+	if (Request.prototype.isPrototypeOf(url)) {
 		request = url;
 	} else {
 		request = new Request(url, options);
@@ -73,8 +73,8 @@ function compileRoute (route) {
   var matchMethod;
   if(method) {
     method = method.toLowerCase();
-    matchMethod = function(options) {
-      var m = options && options.method ? options.method.toLowerCase() : 'get';
+    matchMethod = function(req) {
+      var m = req && req.method ? req.method.toLowerCase() : 'get';
       return m === method;
     };
   } else {
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -1,6 +1,7 @@
 'use strict';
 
 let Headers;
+let Request;
 let Response;
 let stream;
 let Blob;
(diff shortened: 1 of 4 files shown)


src/fetch-mock.js
```diff
@@ -56,6 +57,16 @@ function mockResponse (url, config) {
 	return Promise.resolve(new Response(body, opts));
 }
 
+function toRequest(url, options) {
+	var request;
+	if (Request.prototype.isPrototypeOf(url) && !options) {
+		request = url;
```
wheresrhys: Could you change this to not check for non-existence of options, or to throw a descriptive error if url is a request but options exists

src/fetch-mock.js
```diff
@@ -91,19 +102,22 @@ function compileRoute (route) {
 			debug('constructing starts with string matcher for route: ' + route.name);
 			expectedUrl = expectedUrl.substr(1);
 			route.matcher = function (url, options) {
-				return matchMethod(options) && url.indexOf(expectedUrl) === 0;
+				var req = toRequest(url, options);
+				return matchMethod(req) && req.url.indexOf(expectedUrl) === 0;
```
wheresrhys: could you change the parameter name in matchMethod's definition to be `req` too please

wheresrhys:
Thanks for fixing this. Aside from my comments it looks good

src/fetch-mock.js
vbauerster: ok, I'm not so good at writing error messages, but how about this one: 'you can pass options to the Request constructor, rather than passing along' ?

src/fetch-mock.js
vbauerster: do you mean like this:
```js
route.matcher = function (req, options) { ... }
```

src/fetch-mock.js
wheresrhys: I actually prefer to just remove the ` && !options` and not worry about the error message.

src/fetch-mock.js
wheresrhys: https://github.com/vbauerster/fetch-mock/blob/request-matcher/src/fetch-mock.js#L76 and L77 - just change all the `options` to `req`

vbauerster:
All is done! Thank you for your comments.

wheresrhys:
Thanks for the changes. I'll release once I've tweaked the documentation

PR #48, opened 2015-12-23
Added prebrowserified build

### LINKED ISSUE #47: Cannot use in non-browserified client environment
The following statement on the README is not accurate:
```
Environment doesn't support requiring fetch-mock?

If your client-side code or tests do not use a loader that respects the browser field of package.json use require('fetch-mock/es5/client'). This can also be loaded directly in a script tag as ./node_modules/fetch-mock/es5/client.js (Disclaimer: I haven't actually tried this, but it should work)
```

`es5/client.js` is just a babel compiled version of `src/client.js`, which still attempts to use require. We need another folder (perhaps `browser/`) that contains a browserify-compiled version of `src/client.js`.

rahatarmanahmed:
Now publishes a standalone pre-browserified client build to `/es5/client-browserified.js`, which loads fetch-mock into the `FetchMock` global variable.

Resolves #47

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /node_modules/
 npm-debug.log
 /es5/
+/dist/
--- a/README.md
+++ b/README.md
@@ -134,7 +134,8 @@ it('should make a request', function (done) {
 
 ### Environment doesn't support requiring fetch-mock?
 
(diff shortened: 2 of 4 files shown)


wheresrhys:
Excellent. Thanks for this. The only change I'd suggest is to remove NODE_ENV=production from the browserifying - will make debugging easier for users

rahatarmanahmed:
Done!

wheresrhys:
released as 3.1.0. Thanks again

PR #62, opened 2016-01-27
Use locally installed version of eslint and mocha

danny-andrews:
Running the tests broke for me, because I didn't have eslint installed globally.


wheresrhys:
Thanks for this.

A pro tip is to add `node_modules` to your `$PATH`. Then you can can run using `mocha` even on local installs (useful for development, but perhaps not for makefile)

PR #63, opened 2016-01-27
Add type param to make json stringification opt-in

danny-andrews:
The main goal of this PR is to allow users of this library to stub non-json or text requests. I ran into a problem where I wanted to stub a call with type='arrayBuffer' and I wasn't able to, because the body was being `JSON.stringified`. Let me know if there are any other changes you'd like to see!

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Replaces `fetch()` with a stub which records its calls, grouped by route, and op
 		* `status`: Set the response status (defaut `200`)
 		* `headers`: Set the response headers. (`object`)
 		* `throws`: If this property is present then a `Promise` rejected with the value of `throws` is returned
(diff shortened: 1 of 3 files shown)


wheresrhys:
Thanks for implementing this. Adding blob support has been on my todo list for a while.

I would however rather use a boolean `parseAsJson` which defaults to `true`. 

Pros of this approach would be 
- I can't see that I would ever want the library to have any built in transforms for other data types (the json one is only there because it's overwhelmingly the most common use case)
- It reduces the risk of people being caught out by e.g. returning an object `{title: '1984', type: 'book'}` - I'm pretty sure (will need to revisit the code to check) this would mean they'd get an empty 200 response if `type` is used by fetchmock

If you can tweak the PR to accommodate this I'll merge once it's done

danny-andrews:
Tweaks made.

wheresrhys:
Looks good. Thanks

PR #65, opened 2016-02-17
Update README example error

danfang:
Small issue with `fetchMock.calls().unmatched()` being used instead of `fetchMock.calls().unmatched`


wheresrhys:
thanks for this

Issue #57, opened 2016-01-04
Having some trouble getting this to work within react/redux/webpack application.


deep-c:
Hi,

Sorry if this isn't the right place to put this however im really struggling to get this working. Ive explained the issue I'm having here http://stackoverflow.com/questions/34574682/unable-to-get-fetch-mock-working-with-webpack-and-redux-react.

Im not sure if I am doing something completely wrong but ive tried so many different things and this is my last resort. Any help would be appreciated. 

Thanks

wheresrhys:
From reading your code it looks like you're not interested in analysing calls to fetch, just in returning a mock response - is that right? In that case you could just return `new Response(url, opts)` using code similar to https://github.com/wheresrhys/fetch-mock/blob/master/src/fetch-mock.js#L17 and leave the call to `fetch` out of your implementation. 

I could expose my mockResponse function as a static helper method if that'd be useful

deep-c:
Hi,

Yea thats right.

I managed to get this working by using window.fetch instead of just fetch. From what i thought, fetch (from isomorphic fetch) for the browser would use that from the global window object which it patches if necessary (think i read that in the docs somewhere). It seems that fetch being referenced in my code isnt the same fetch as that of the browser window object (which upon logging in the console returns the mock function). 

This just means that i have to annoyingly reference window.fetch until i get my backends up and then change all references to fetch when i dont need the mock response any longer and I need to start rendered things server side.

SpaceviewMob:
We had the exact same issue. Found a solution, but it required a bit of extra code that we have now shared.

https://github.com/spaceviewinc/fetch-mock-forwarder

wheresrhys:
I'd happily accept  PR to document this in the troubleshooting section of the readme

theosherry:
Hey, I was having similar problems, and I resolved it by importing isomorphic-fetch to the global namespace rather than importing it to a variable.  I see in your code that you were importing it to a variable as well, so this might be a solution for you.

esBeee:
Hi!
I am having a very similar issue, also with a react/redux/webpack application, and am running out of ideas. I try to use fetch-mock in a unit-test, that's supposed to test a function I import from another file, but a real request is being performed (I've also described the problem more exact [on StackOverflow](http://stackoverflow.com/questions/35414966/mocking-an-outgoing-request-inside-an-async-redux-action-in-a-node-app-doesnt-w)). Can anyone please give me a hint on what I am missing?

[...]

PR #67, opened 2016-02-26
Document usage with ES6 modules syntax

MattiSG:
Related to #57.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,9 +127,9 @@ it('should make a request', function (done) {
 ## Troubleshooting
 
 ### `import` syntax (ES6 modules)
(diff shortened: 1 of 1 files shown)


wheresrhys:
Thanks for documenting this (been waiting for a while for someone actually affected by the issue to help me out with it). 

That said, as isomorphic-fetch sets fetch as a global shouldn't `import 'isomorphic-fetch'` work? If it doesn't for you, could you tell me more about why not.

MattiSG:
Indeed, it does work! I simply did not know about this syntax… Thanks!

MattiSG:
Updated PR accordingly  :)

wheresrhys:
Sorry for slow response and thanks for the update. I have a feeling this issue affects projects using require too, so will add a note after merging

Issue #70, opened 2016-03-14
Mockery doesn't work as expected


timnew:
I setup the `mockery` and `fetch-mock` as following code:

```js
import mockery from 'mockery'
import fetch from 'node-fetch'
import fetchMock from 'fetch-mock'
import MyModule from './MyModule'

fetchMock.useNonGlobalFetch(fetch)
// ...
    beforeEach(() => {
      mockery.enable({ useCleanCache: true })
      mockery.registerMock('node-fetch', fetchMock.mock(pattern, 'POST', 200).getMock())
    })

    afterEach(() => {
      mockery.deregisterMock('node-fetch')
      mockery.disable()
    })
```

And during the test, I found the request doesn't get mocked as expected, all request hit the real server and `fetchMock.called()` returns false

Did I missed something or what should I do?!?!

wheresrhys:
You need to import `MyModule` after calling `mockery.registerMock`, which you can only do if using `require('./MyModule')`. I don't know if there are any other libraries that mock modules that deal with `import` better than mockery does. I doubt it though because they'd need to pretty closely coupled with babel

timnew:
I realized the issue, and trying to figure out a solution. 
It is kind of painful while writing test for library rather than application.
I think the only way to solve the issue is to add another abstraction layer in the lib to enable developer to inject fetch ployfill into it.
Thanks for the help anyway

Primajin:
I'm also having the Problem, that fetchMock does not actually mock away my request, and requests are running against real url.
I am using `import fetch from 'isomorphic-fetch';` in my react action.
I cant use mockery since it does not work with Webpack (https://github.com/mfncooper/mockery/issues/37). Do you have any different Idea?

wheresrhys:
try `import  'isomorphic-fetch'` instead of `import fetch from 'isomorphic-fetch';`

Primajin:
Great! I then first ran into #60 but with adding `require('isomorphic-fetch');` before my actual tests (e.g. `loadtests.js`) now everything runs smooth. You saved my day (and my code) Thanks!!

shivasai09:
hi there, 
i am currently using fetch API to get the data from the server so i am using it like this
callApi.js
```
import fetch from 'isomorphic-fetch';
class CallApi {
static getInfo(){
fetch(url,body).then(res=>{res.json()})
}
}
export default CallApi
```
and in my  other file `book.js` i am using like this 
```
import CallApi from (./callApi.js)
...
[...]

PR #72, opened 2016-03-17
Update READ.me to make part of mockery more clear

timnew:
Kind of related to issue #70 

* Register mock `node-fetch` instead of `fetch`, since `node-fetch` is used in the code, which is more straightforward.
* Move `deregisterMock` call to `afterEach` hook, since current implementation might cause mock leak if test code failed
* Add the instruction to use `require` rather than `import`

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ afterEach(function(done) {
 })
 ```
 
-**HINT:** You need to user `require` instead of `import` to require the you modoule, since `mockery` strongly depends the order of execution. To mock `fetch` properly, `registerMock` must happens before `require` the module. And `import` will be always executed as the very first statements of the file.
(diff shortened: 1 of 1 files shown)


wheresrhys:
I've just merged some biggish refactors and changes to documentation, including a lot more on use with mockery. I think all your points are addressed, but feel free to open another PR if you can think of any improvements

timnew:
OK 


Cheers,

TimNew

Geek | FullStack Developer | Mobile Developer | NightTime Photographer | Hardware Amateur
Blog:        timnew.me (http://timnew.me)
Github:    github.com/timnew (http://github.com/timnew)
LinkedIn: linkedin.com/in/timwen (http://linkedin.com/in/timwen)
Twitter:    @timnew (http://twitter.com/timnew)

timnew:
I still think current implementation might cause mock leak when test failed.

Since when assertion failed, the `mockery.deregisterMock('node-fetch')` will never get executed.
An easy and safe way to solve the issue is to move this call into `afterEach` hook, which is guaranteed to be executed after every test, whether it is passed or not.

And there is a typo: `mockery.deregisterMock('fetch')` should be `mockery.deregisterMock('node-fetch')`

Issue #59, opened 2016-01-13
Add extra params in response object


k010:
Hey guys, thanks for this is great !!
It would be nice to add some extra param in response object for eg: *statusText*, *etc*... [Fetch Response](https://developer.mozilla.org/en-US/docs/Web/API/Response)

What do you think ?
Tks !

wheresrhys:
statusText is the only one missing as far as I can see https://developer.mozilla.org/en-US/docs/Web/API/Response/Response. For completeness, yes that should be configurable.

k010:
:+1:  Thanks !!

PR #73, opened 2016-03-17
Add statusText to mocked response

timnew:
For issue #59


timnew:
@wheresrhy I found you use cascaded callbacks in the test, which might hang the test if test failed, since done is never called then.

Actually `mocha` has the built-in support to promise, so you don't need to invoke `done` manually.

And you can always return a promise in the `then` callback, which will be resolved automatically, so you don't need to use the fragile cascaded callbacks.

wheresrhys:
Thanks for this. Would you mind making a few stylistic changes

- rename statusText.js to status-text.js
- rename STATUS_TEXT to statusTextMap

timnew:
Fixed, please check whether it is okay now.

wheresrhys:
Yep looks good. Just needs conflicts resolved (hopefully won't be too difficult)

timnew:
I should be easy I think 😁

wheresrhys:
Have merged for you in another PR. Thanks for the contribution

PR #84, opened 2016-04-11
Use promises instead of done() for async tests

motiz88:
This uses Mocha's neat [promise support](https://mochajs.org/#working-with-promises) and a bit of promise chaining to generate more meaningful test failures during development while keeping the code clean.

(The alternative would have been to put a `try { ... } catch(e) { done(e); }` around each deferred `expect` block, and that's just plain nasty)

I'll happily rebase #83 in this style if this PR is accepted.


wheresrhys:
great - thanks

PR #116, opened 2016-07-20
Replace Object.assign with an inline helper when transpiling to ES5

### LINKED ISSUE #115: v5 fails on Node < 4 due to use of Object.assign
As of 5.x, using `fetch-mock/es5/server` on Node 0.10, 0.12, or iojs throws the following error:

```
TypeError: Object function Object() { [native code] } has no method 'assign'
```

wKovacs64:
Fixes #115.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,5 @@
+{
+  "plugins": [
+    "transform-object-assign"
+  ]
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,0 @@
-{
-  "plugins": [
-    "transform-object-assign"
-  ]
-}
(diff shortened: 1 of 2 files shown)


wheresrhys:
Thanks for this. I would rather the transform got called from within the npm script though https://github.com/wheresrhys/fetch-mock/blob/master/package.json#L10, and that there was no .babelrc added

wKovacs64:
Updated. If that's what you had in mind, let me know if you want them squashed before merging.

wheresrhys:
Yep perfect thanks. Gonna merge as the failing tests relate to an upstream dependency of a new feature I'm waiting for a release to be made for

PR #121, opened 2016-08-09
Add a basic usage example

Wilfred:
fetch-mock has examples for using with mockery, and configuring, but I struggled with getting the basics working.

I've written a self-contained example that shows basic usage, to help new users.


Wilfred:
Test failures are unrelated -- I've only changed the README.

wheresrhys:
Thanks. The test failures are now fixed.

I'm gonna mull over whether I'd like this above or below the API docs before merging, but other than that thanks for putting the effort into a really clear example.

PR #123, opened 2016-08-17
Remove browserify field from package.json and change build/publish path

iamolivinius:
I was unable to use this package in my environment. My setup is based on browserify and Typescript. When I tried to bundle my test files where I required('fetch-mock') browserify looked up the transform field and tried to load babelify. As soon as I removed the browserify field from package.json everything works fine. 

To compensate the changes a added the commonly used .babelrc file with the same settings as in package.json and added the babelify transform to the browserify command line arguments and in the karma config. 

Furthermore I changed the main field to point to the already transpiled server.js file in the es5 folder.

Next I added the files field to package.json. This lists all files that should be included when publishing the package to npm. I don't think you need to publish more then the *.md and the transpiled (& bundled) *.js files.

And for ease of use during dev I added also a npm run build command.

Tests: Chrome 52.0.2743 (Linux 0.0.0): Executed 61 of 61 SUCCESS (0.191 secs / 0.089 secs)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,8 @@
+{
+  "presets": [
+    "es2015"
+  ],
+  "plugins": [
+    "transform-object-assign"
+  ]
+}
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -12,7 +12,8 @@ module.exports = function(karma) {
 			'test/client.js': ['browserify']
 		},
 		browserify: {
-			debug: true
+			debug: true,
+			transform: ['babelify']
 		},
 		browsers: ['Chrome'],
 		customLaunchers: {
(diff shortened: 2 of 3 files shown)


amit1911:
+1

wheresrhys:
I'm a bit nervous about merging this because I've experienced problems in the past where a dependency's inclusion of a .babelrc file that isn't compatible with my project's version of babel has broken the install/build. Potentially it's possible to get the es5 bundle built and the tests to pass without any config json anywhere, which I'll investigate.

The decision to point at src/server.js is deliberate - the transpiled files are there to support node 0.x, but I think it's a better developer experience in node 4 and above if the original un-transpiled files are used.

So I'm going to reject this, but will try to find another solution to the problems you've experienced

wheresrhys:
Let me know if v5.1.1 doesn't fix your problems

iamolivinius:
This works for me. Thanks for the quick response :heart:

PR #131, opened 2016-09-16
Browserify es5/client.js, not src/client.js

### LINKED ISSUE #130: es5/client-browserified.js actually not transpiled into ES5
I'm running `fetch-mock` as part of a workflow involving `karma` and `phantomJS`, and just hit an error like:

```
SyntaxError: Unexpected token 'const'
  at node_modules/fetch-mock/es5/client-browserified.js:4
```

That file has, indeed, ES6-only features like `const` and `let` and seems like it's not being transpiled properly.

I think the root cause is running `browserify` with `src/client.js` as the entry point, instead of `es5/client.js` as the entry point (this is in the `scripts` section of `package.json`).

IvanSanchez:
Fixes #130

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "browser": "es5/client.js",
   "scripts": {
     "test": "make test",
-    "browserify": "browserify -s fetchMock src/client.js > es5/client-browserified.js",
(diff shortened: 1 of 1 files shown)


wheresrhys:
Thanks for this. Will republish now

PR #133, opened 2016-09-17
Fixes bug in interaction with babel-transform-runtime - Fixes #132

### LINKED ISSUE #132: Returning Promise in older browsers
Hello together,

I'm using fetch-mock to create a fake-backend for development. That's why every request fails with a specific probability and gets resolved after x ms. In chrome this works fine, but in IE and Firefox, it doesn't. I searched through the fetch-mock source a bit and noticed, that this line:

`if (response instanceof Promise) {`

is always returning false in Firefox and in IE.

Update: I investigated some more and i found out that `Promise` in the fetchMocks source is not the same as `Promise` in my source... I guess some Babel or webpack vodoo is going on there :x

Update 2: I could fix this by removing the babel-runtime plugin from my webpack config. Still not sure what is really causing this...

Update 3: Okay I think I understand this now: `babel-plugin-transform-runtime` replaces every call to built in ES6 types (like Promise) with a reference to the babel-core polyfill, so it does NOT polyfill the globals. Since babel should not convert node_modules/ it cannot convert the Promise inside of fetchMock, so that this Promise and babel promise don't share the same identity.

andreasgruenh:
Removes reference to Promise builtin so that check for promise works
even when the Project is using a non global Promise polyfill
(Like what happens with babel-transform-runtime)

### DIFF
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -241,7 +241,8 @@ class FetchMock {
 			response = response (url, opts);
 		}
 
-		if (response instanceof Promise) {
+		// Checks if the response is a promise without using Promise builtin
+		if (typeof response.then === 'function') {
 			return response.then(response => this.mockResponse(url, response, opts))
 		} else {
 			return this.mockResponse(url, response, opts)


andreasgruenh:
I can't seem to get the tests running on my computer :( npm run test always fails.

wheresrhys:
It should be `make test`

wheresrhys:
Closed in favour of https://github.com/wheresrhys/fetch-mock/pull/137, but thanks for contributing

Issue #117, opened 2016-07-21
Feature request: support glob expressions


capaj:
It would be very nice if I could write the matcher as glob expression string. Glob expressions are much easier to write and parse then regexes and  are perfect for this usecase.

wheresrhys:
Good idea. I also think expressjs style path syntax would be a useful addition.

That said, it's difficult to know how to elegantly add support for these to the API as both globs and express path expressions are just strings. 

1. duck-typing the strings, but I doubt this can be made robust
2. pass in an option in the third parameter, but I don't like the verbosity
3. as above, but able to set a global option to treat the strings as globs/express paths/urls by default
4. More prefixes e.g. currently `^` indicates the url must begin with a string. For backwards compatibility this could still be supported, but deprecated and superceded by

wheresrhys:
done

PR #145, opened 2016-10-02
beginning work on adding support for glob, path, end and begin

wheresrhys:


wheresrhys:
fixes https://github.com/wheresrhys/fetch-mock/issues/117

PR #157, opened 2016-12-16
Allow passing a custom promise implementation to sandbox()

### LINKED ISSUE #78: Support swapping out Promise library
I'm trying to use `fetch-mock` to against environments which do not have `Promise` implemented natively.

I'm using `isomorphic-fetch` and replacing its Promise implementation with bluebird, so it works in these environments in my production code.

I'm using it something like this:

```
import 'isomorphic-fetch'
import Promise from 'bluebird'
fetch.Promise = Promise
```

It would be cool if `fetch-mock` used a similar mechanism, or (perhaps even better?) looked into the original `fetch` and used its `.Promise` property if it exists.

Let me know if you want me to try to PR this (I assume it would be pretty quick and self-evident how to do this)

pimterry:
This fixes #78.

### DIFF
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -128,7 +128,7 @@ FetchMock.prototype.addRoute = function (route) {
 
 
 FetchMock.prototype.mockResponse = function (url, responseConfig, fetchOpts) {
-	let Promise = this.fetchMock.promise || FetchMock.global.Promise;
+	let Promise = this.fetchMock.Promise || FetchMock.global.Promise;
 
 	// It seems odd to call this in here even though it's already called within fetchMock
 	// It's to handle the fact that because we want to support making it very easy to add a
@@ -271,7 +271,7 @@ FetchMock.setGlobals = function (globals) {
 	Object.assign(FetchMock, globals)
 }
 
-FetchMock.prototype.sandbox = function (promise) {
+FetchMock.prototype.sandbox = function (Promise) {
 	if (this.routes.length || this.fallbackResponse) {
 		throw new Error('.sandbox() can only be called on fetch-mock instances that don\'t have routes configured already')
 	}
@@ -283,7 +283,7 @@ FetchMock.prototype.sandbox = function (promise) {
 	);
 	instance.fetchMock.bindMethods();
 	instance.fetchMock.isSandbox = true;
-	instance.fetchMock.promise = promise;
+	instance.fetchMock.Promise = Promise;
 	this.restore();
 	return instance.fetchMock;
 };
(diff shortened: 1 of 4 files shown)


pimterry:
Build has failed, but it passes locally for me. Looks like the same spurious failure as https://travis-ci.org/wheresrhys/fetch-mock/builds/183724306, which suggests it's just an intermittent issue with your karma tests? Let me know if that doesn't sound plausible and I'll investigate further.

src/fetch-mock.js
```diff
@@ -269,7 +271,7 @@ FetchMock.setGlobals = function (globals) {
 	Object.assign(FetchMock, globals)
 }
 
-FetchMock.prototype.sandbox = function () {
+FetchMock.prototype.sandbox = function (promise) {
```
emirotin: I'd call it `Promise` with first capital to emphasize it's a constructor, not an instance (and same everywhere where it's passed around).

pimterry: Good point, done.

pimterry:
Super conveniently, that extra change has rerun the build, so it now passes 😄

emirotin:
😎

test/spec.js
```diff
@@ -1092,6 +1092,15 @@ module.exports = (fetchMock, theGlobal, Request, Response) => {
...
 			});
 
+			it('takes an optional Promise to use', () => {
+				const BluebirdPromise = require('bluebird');
+				const sbx = fetchMock
+					.sandbox(BluebirdPromise)
+					.mock('http://example.com', 200);
+
+				expect(sbx('http://example.com')).to.be.instanceof(BluebirdPromise);
+			});
```
wheresrhys: Could you add an additional test(s0 to make sure sandboxed instances don't mess up promise references for other instances or the global fetch-mock instance

pimterry:
Sure, done.

wheresrhys:
Thanks for adding the extra tests. I was actually thinking of something slightly more than this though. You mentioned that you depend on some of Bluebird's non-standardised extensions to the promise object. In this case, wouldn't you want to be sure that fetch-mock always returns a Bluebird Promise rather than a native promise? To verify this I think each test should also check whether the promises returned are instances of Bluebird/GlobalPromise

pimterry:
@wheresrhys Sure, done.

This was a good catch actually - there was a bug (or at least, some poorly defined behaviour) where if you configured the response using a different promise implementation, that's the promise implementation that would eventually be returned later from the fetch call. I've now made this fully consistent: for any fetch instance, you'll always get the same type of promise back, either the global one or the one locally configured in your sandbox.

Tests seem to have failed spuriously in the browser again. Pretty sure they'll work if you rerun them, and they do all pass for me locally.

wheresrhys:
Yep all good. Thanks for working on this. I'll release a minor version in a few minutes

wheresrhys:
released as 5.8.0

pimterry:
Fantastic, thanks!

Issue #159, opened 2016-12-20
The Fetch mock returned by sandbox breaks if restore() is called


pimterry:
```javascript
var fetch = require('fetch-mock').sandbox();

fetch.get('https://api.resin.io/foo', 200);
fetch('https://api.resin.io/foo').then((response) => {
	console.log(response.status);  // Prints 200, great.

	fetch.restore();
	fetch.get('https://api.resin.io/foo', 500);
	return fetch('https://api.resin.io/foo');
}).then((response) => console.log(response.status)) // Still prints 200!
```

This prints 200 twice, which is very surprising. It seems that `restore()` basically breaks future mocking but keeps all the current set of mock routes.

You can work around this with:

```javascript
var fetchMock = require('fetch-mock').sandbox();
var fetch = fetchMock.fetchMock;

fetchMock.get('https://api.resin.io/foo', 200);
fetch('https://api.resin.io/foo').then((response) => {
	console.log(response.status);  // Prints 200, great.

	fetchMock.restore();
	fetchMock.get('https://api.resin.io/foo', 500);
	return fetch('https://api.resin.io/foo');
}).then((response) => console.log(response.status)) // Now prints 500
```

I.e. using `sandbox().fetchMock` as your fetch stub, not the returned function directly, but keeping the originally returned `sandbox()` value to call mock methods on.

I think this isn't what's supposed to happen - the code and docs suggest that the sandbox returns a fake fetch that you can just start using. This at least should get documented, but should probably be changed to work as expected above, imo.

pimterry:
Sorry to spam so many issues in quick succession. Hopefully some of these are useful!

wheresrhys:
Yep, very useful. It's a new feature I don't use myself, so bound to be edge cases I hadn't thought of. I'll write some tests and hopefully be able to fix it. The worst case scenario is I may have to document sandboxed fetch-mocks as single use constructs on which `restore` cannot be called

pimterry:
Ah, sorry a more problematic bonus point here: this breaks #157 (the custom promises PR from yesterday).

`require('fetch-mock').sandbox(BluebirdPromise)("http://google.com")` does return a custom promise correctly, and everything's great. It breaks `restore()` though, as above, so you need a workaround.

`require('fetch-mock').sandbox(BluebirdPromise).fetchMock("http://google.com")` does work with restore(), but has a different `this`, so doesn't find `this.fetchMock.Promise`, and just returns native promises. Failing test:

```javascript
[...]

PR #160, opened 2016-12-20
fixing problems with sandboxed fetchMock being incorrectly bound

wheresrhys:
@pimterry

### DIFF
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -80,7 +80,7 @@ FetchMock.prototype.spy = function () {
 }
 
 FetchMock.prototype.fetchMock = function (url, opts) {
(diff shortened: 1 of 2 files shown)


wheresrhys:
Should fix https://github.com/wheresrhys/fetch-mock/issues/159... though I now wonder if the convenience of having `sandbox()` return a function justifies the hackery needed to get it to work

pimterry:
Back from Christmas now. Thanks for this, looks great, and seems to neatly solve #159 in our codebase.

Personally I quite like the resulting sandbox API this gives, so if I were you I'd stick with it, but I can see how the fiddly complexity might be annoying. Seems to work nicely now either way.

Issue #158, opened 2016-12-20
Fetch-mock in the browser requires a global fetch implementation


pimterry:
In the browser code `fetch-mock` assumes that Request, Response and Header globals are available, which means non-global development with fetch isn't possible with fetch-mock in browsers.

Our current workaround is to just expose all these globals in the tests:

```javascript
if (IS_BROWSER) {
	realFetchModule = require('fetch-ponyfill')({ Promise: BluebirdPromise })
	global.Headers = realFetchModule.Headers
	global.Request = realFetchModule.Request
	global.Response = realFetchModule.Response
}
```

It'd be nice to not have to do that, since it's a little messy, and because this clearly isn't necessary, as the Node implementation doesn't make these assumptions and works fine without these globals.

This would be a fairly niche problem, except that PhantomJS doesn't support fetch, and that's where most browser automated testing happens.

wheresrhys:
My issue with this is that `fetch` and its associated classes are supposed to be global in the browser https://fetch.spec.whatwg.org/, so its a pretty reasonable assumption that the user's app takes care of making these globals available before trying to run either tests or application code. However one thing I'm increasingly aware of is that the use cases for fetch-mock are quite varied - and ponyfills are a lot more in vogue now than a few months ago - so I'm open to making your use case easier, but with the proviso that the default behaviour stays as it is.

Do you have any suggestions? Is there any reason you can't use one of these approaches https://github.com/wheresrhys/fetch-mock#polyfilling-fetch

pimterry:
The main problem for us is that we're trying to provide modules that will be used in other people's code. We don't want to write new globals for them unexpectedly (e.g. by including isomorphic-fetch automatically), but we'd also like to make it really easy to pick up and use these libraries (so I'd like to avoid 'Install our library, and also remember to install and require isomorphic-fetch yourself in advance if you're in Node or an old browser' instructions everywhere).

We're building a library, but we're not developing the end user application, so we can't dictate what exists in the resulting environment.

We could polyfill the environment more actively for our tests (and that's basically what we're doing in the workaround above), it's just that it'd be nice to test an environment that's more representative of our end user case. They've only imported the library, and they may or may not have fetch available, and our library should Just Work. This isn't desperately urgent or anything, you can work around it, but it will be a stumbling block for anybody else doing this too.

In terms of solutions, I'd suggest tying this into sandbox again to keep the existing behaviour normally, so you can write:

```javascript
[...]

PR #161, opened 2017-01-15
added setImplementations method to use custom fetch utility classes

wheresrhys:
@pimterry it'd be good if you could try this out instead of the hack mentioned in https://github.com/wheresrhys/fetch-mock/issues/158. The tests may not comprehensively test all the paths where Request/Response/Headers are used, so I will add more tests if you uncover any failures when using in your tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,9 @@ Returns the options for the last matched call to fetch
 Set some global config options, which include
 * `sendAsJson` [default `true`] - by default fetchMock will convert objects to JSON before sending. This is overrideable fro each call but for some scenarios e.g. when dealing with a lot of array buffers, it can be useful to default to `false`
 
+##### `setImplementations(opts)`
+When using non global fetch (e.g. a ponyfill) or an alternatice Promise implementation, this will configure fetch-mock to use your chosen implementations. `opts` is an object with one or more of the following properties: `Headers`,`Request`,`Response`,`Promise`. Note that `setImplementations(require('fetch-ponyfill'))` will configure fetch-mock to use all of fetch-ponyfill's classes. 
+
 ## Troubleshooting and alternative installation
 
 ### `fetch` is assigned to a local variable, not a global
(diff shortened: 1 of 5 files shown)


pimterry:
Sorry I didn't get around to testing this very quickly - it's been a fairly hectic start to the year.

I now have updated some of our code to use this, but there's a large problem, and a much smaller problem.

**Large problem**: in this PR `client.js` changes to include `Promise` directly by name, which means in any environment without promises (e.g. phantomjs) loading `fetch-mock` immediately throws "Can't find variable: Promise" and explodes. It seems to work if you move to `theGlobal.Promise`, and I've had a quick look at putting in a PR, but the current master tests actually fail for me locally in Chrome, which is surprising! I'll take a closer look next week - if you have any suggestions that would be helpful. I see lots of:

```
Chrome 56.0.2924 (Mac OS X 10.11.6) fetch-mock configurability fetch utility class implementations should use configured Headers FAILED
TypeError: Failed to construct 'Headers': Please use the 'new' operator, this DOM object constructor cannot be called as a function.
```

**Much smaller problem**: the setImplementations api is a bit awkward when you want to set both Promise and the fetch implementations. Our usage with these changes right now looks like:

```javascript
var _ = require('lodash')
var Promise = require('bluebird')
var realFetch = require('fetch-ponyfill')({ Promise: Promise })
fetchMock.setImplementations(_.assign({ Promise: Promise}, realFetch))
```

There's an awkward _.assign required to combine the fetch-ponyfill exports with promises, if you're using both (and the api does explicitly support using both).

As a related aside, `setImplementations(require('fetch-ponyfill'))` (from the docs) doesn't actually work normally anyway, as `fetch-ponyfill` returns a constructor, but the reality is *pretty* close (if you don't override promises): `setImplementations(require('fetch-ponyfill')())`.

If you are overriding promises though, I think the above is unavoidable right now. We could improve this? Something like the below should be possible, without breaking the API:

```javascript
fetchMock.setImplementations(require('fetch-ponyfill')({ Promise: Promise }), Promise)
```

wheresrhys:
Could you raise a bug for the first issue. Should probably start running tests in phantom too.

I'll fix the docs now, but I'm reasonably happy with the API. It's a one time setup thing before running tests, so I don't think the awkwardness of use is too big a problem as the API is simpler conceptually - 'pass in an object of the Constructors you want to use'

pimterry:
Ok, keeping the api as is sounds reasonable. I've filed separate bugs for the [failing tests](https://github.com/wheresrhys/fetch-mock/pull/167) and the [global promise](https://github.com/wheresrhys/fetch-mock/pull/166) issue.

PR #186, opened 2017-05-12
Add failing test cases for usage of custom headers

winterstefan:
Hi there,

I had some strange handling with `fetch-mock` and testing custom header definitions.
Therefore, I opened this PR by adding two test cases, that will fail (tested locally).

It would be great if you could check this PR and leave me a reply on it. (Please check the two comments for detailed information.)
_IMHO_ the `getHeaderMatcher` in [compile-route.js](https://github.com/wheresrhys/fetch-mock/blob/master/src/compile-route.js#L22) should / could be enhanced to handle the two cases.

Don't hesitate contacting me if you have any question on it.

### DIFF
--- a/src/compile-route.js
+++ b/src/compile-route.js
@@ -1,6 +1,7 @@
 'use strict';
 const glob = require('glob-to-regexp')
 const express = require('path-to-regexp');
+const Headers = require('node-fetch').Headers;
 
 const stringMatchers = {
 	begin: targetString => {
@@ -27,6 +28,11 @@ function getHeaderMatcher (expectedHeaders) {
 		if (!headers) {
 			headers = {};
 		}
+
+		if (headers instanceof Headers) {
+			headers = headers.raw();
+		}
+
 		const lowerCaseHeaders = Object.keys(headers).reduce((obj, k) => {
 			obj[k.toLowerCase()] = headers[k]
 			return obj;
--- a/src/compile-route.js
+++ b/src/compile-route.js
@@ -37,10 +37,37 @@ function getHeaderMatcher (expectedHeaders) {
 			obj[k.toLowerCase()] = headers[k]
 			return obj;
 		}, {});
+
 		return expectation.every(header => {
-			return lowerCaseHeaders[header.key] === header.val;
+			return areHeadersEqual(lowerCaseHeaders, header);
 		})
 	}
+}
+
+function areHeadersEqual (currentHeader, expectedHeader) {
+    const {key, val} = expectedHeader;
+    const currentHeaderValue = (Array.isArray(currentHeader[key])) ? currentHeader[key] : [currentHeader[key]];
+    const expectedHeaderValue = (Array.isArray(val)) ? val : [val];
+
+    if (currentHeaderValue === expectedHeaderValue) {
+		return true;
+    }
+
+    if (!currentHeaderValue || !expectedHeaderValue) {
+		return false;
+	}
+
+    if (currentHeaderValue.length !== expectedHeaderValue.length) {
+		return false;
+    }
+
+    for (let i = 0; i < currentHeaderValue.length; ++i) {
+        if (currentHeaderValue[i] !== expectedHeaderValue[i]) {
+			return false;
+        }
+    }
+
+    return true;
 }
 
 function normalizeRequest (url, options, Request) {
--- a/src/compile-route.js
+++ b/src/compile-route.js
@@ -45,7 +45,8 @@ function getHeaderMatcher (expectedHeaders) {
 }
 
 function areHeadersEqual (currentHeader, expectedHeader) {
-    const {key, val} = expectedHeader;
+    const key = expectedHeader.key;
+    const val = expectedHeader.val;
     const currentHeaderValue = (Array.isArray(currentHeader[key])) ? currentHeader[key] : [currentHeader[key]];
     const expectedHeaderValue = (Array.isArray(val)) ? val : [val];
 
--- a/src/compile-route.js
+++ b/src/compile-route.js
@@ -50,14 +50,6 @@ function areHeadersEqual (currentHeader, expectedHeader) {
     const currentHeaderValue = (Array.isArray(currentHeader[key])) ? currentHeader[key] : [currentHeader[key]];
     const expectedHeaderValue = (Array.isArray(val)) ? val : [val];
 
-    if (currentHeaderValue === expectedHeaderValue) {
-		return true;
-    }
-
-    if (!currentHeaderValue || !expectedHeaderValue) {
-		return false;
-	}
-
     if (currentHeaderValue.length !== expectedHeaderValue.length) {
 		return false;
     }
--- a/test/custom-header.spec.js
+++ b/test/custom-header.spec.js
@@ -1,0 +1,64 @@
+'use strict';
+const expect = require('chai').expect;
+
+module.exports = (fetchMock, theGlobal, Headers) => {
+
+    describe('fetch-mock', () => {
+
+        const dummyFetch = () => Promise.resolve(arguments);
+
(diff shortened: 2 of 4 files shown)


test/custom-header.spec.js
```diff
@@ -0,0 +1,64 @@
...
+                    name: 'route1',
+                    headers: {
+                        test: 'yes'
+                    },
+                    matcher: 'http://it.at.here/',
+                    response: 'ok'
+                }).catch();
+                return Promise.all([
+                    fetch('http://it.at.here/', {headers: {test: 'yes'}}),
+                    fetch('http://it.at.here/', {headers: new Headers({test: 'yes'})}),
```
winterstefan: **Case 1**

In one of our projects we're using fetch with the `Headers` class from [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch). 

_Example API call_
```javascript_
fetch('/api', {
    'method': 'POST',
    'headers': new Headers({'X-CHUCK': 'NORRIS'}),
})
```

_Example test call_
```javascript
fetchMock.postOnce('/api', new Response(stubResponse, { status: 200 }), {headers: {'x-chuck': 'norris'}});
```

Without the third parameter, this mock will match and the test succeeds. But passing the custom headers param will result in an **unmatched api call**.
Internally, the `X-CHUCK` header is stored as array:
```javascript
console.log(fetchMock.lastCall()[1].headers._headers);
// { 'x-chuck': ['NORRIS'] }
```

**Question**: Is there any way using the `Headers` class that way? What's your suggestion on it?

test/custom-header.spec.js
```diff
@@ -0,0 +1,64 @@
...
+                    name: 'route1',
+                    headers: {
+                        test: ['foo', 'bar']
+                    },
+                    matcher: 'http://it.at.here/',
+                    response: 'ok'
+                }).catch();
+                return Promise.all([
+                    fetch('http://it.at.here/', {headers: {test: 'yes'}}),
+                    fetch('http://it.at.here/', {headers: {test: ['foo', 'bar']}}),
```
winterstefan: **Case 2**

Case 1 led me to a general question. Even if not using the mentioned `Headers` class, a custom HTTP header may have _n_ values (see section 4.2 of [RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html)).

_Example API call_
```javascript
// With Headers class
const headers = new Headers();
headers.append('X-WEBMASTER', 'CHUCK NORRIS');
headers.append('X-WEBMASTER', 'BATMAN');

fetch('/api', {
    'method': 'POST',
    'headers': headers,
})

// With plain notation
fetch('/api', {
    'method': 'POST',
    'headers': {'X-WEBMASTER': ['CHUCK NORRIS', 'BATMAN']},
})

```

**Question**: Is there a way of testing a HTTP request with such header setup with `fetch-mock`? What's your suggestion on it?

wheresrhys:
Thank for submitting this. Both seem like a totally valid use cases - I aim for fetch-mock to support any valid call of fetch. 

I'll try to look at fixing them. Could you skip the tests in your PR so I can merge into master. Or if you feel like fixing the bugs yourself be my guest

winterstefan:
@wheresrhys, thank you for your very quick response!
I commented out the `custom-header.spec`, so all PR _checks have been passed_.

I'll give it a try fixing these two cases (maybe tomorrow) and let you know in a PR.

src/compile-route.js
```diff
@@ -27,16 +28,48 @@ function getHeaderMatcher (expectedHeaders) {
...
+
+    if (!currentHeaderValue || !expectedHeaderValue) {
+		return false;
+	}
+
+    if (currentHeaderValue.length !== expectedHeaderValue.length) {
+		return false;
+    }
+
+    for (let i = 0; i < currentHeaderValue.length; ++i) {
```
wheresrhys: I wonder if the order of items in the array should matter? Currently it does. Should both arrays be sorted before comparing?

src/compile-route.js
```diff
@@ -27,16 +28,48 @@ function getHeaderMatcher (expectedHeaders) {
...
+function areHeadersEqual (currentHeader, expectedHeader) {
+    const {key, val} = expectedHeader;
+    const currentHeaderValue = (Array.isArray(currentHeader[key])) ? currentHeader[key] : [currentHeader[key]];
+    const expectedHeaderValue = (Array.isArray(val)) ? val : [val];
+
+    if (currentHeaderValue === expectedHeaderValue) {
+		return true;
+    }
+
+    if (!currentHeaderValue || !expectedHeaderValue) {
```
wheresrhys: is `!expectedHeaderValue` a case we should care about here?

src/compile-route.js
```diff
@@ -27,16 +28,48 @@ function getHeaderMatcher (expectedHeaders) {
...
+
 		return expectation.every(header => {
-			return lowerCaseHeaders[header.key] === header.val;
+			return areHeadersEqual(lowerCaseHeaders, header);
 		})
 	}
 }
 
+function areHeadersEqual (currentHeader, expectedHeader) {
+    const {key, val} = expectedHeader;
```
wheresrhys: This fails in CI as untranspiled fetch-mock supports node4.

src/compile-route.js
winterstefan: In my humble opinion, the order of the values (multiple values per single header field) **do** matter. Therefore I decided to do a strict comparison.
According to the [RFC 2616, sec. 4.2](https://tools.ietf.org/html/rfc2616#section-4.2), the order in between both notations (comma separated list vs. multiple headers with identical name) is significant and no interceptor / proxy is allowed to change this order.

>    It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma. The order in which header fields with the same
   field-name are received is therefore significant to the
   interpretation of the combined field value, and thus a proxy MUST NOT
   change the order of these field values when a message is forwarded.

In assumption of that, I'd recommend **not** sorting before the comparison. What do you think?

src/compile-route.js
winterstefan: Sry, my bad. I'll change that.

src/compile-route.js
winterstefan: I just tried to be as much error-prone as possible. You're correct - I cannot think of a way of having an invalid `expectedHeaderValue` in here.

In fact, because of wrapping both incoming params into an array, the usage of `(currentHeaderValue.length !== expectedHeaderValue.length)` should be enough, shouldn't it?

src/compile-route.js
wheresrhys: yep if the spec says it's significant it's definitely the right choice

src/compile-route.js
wheresrhys: yep, and come to think of it the check for strict equality on the lines above would always fail too for the same resaon, so can be removed too

test/spec.js
```diff
@@ -1442,5 +1442,9 @@ e.g. {"body": {"status: "registered"}}`);
 
 		})
 
+        afterEach(() => {
```
wheresrhys: would you mind trying to remove the `beforeEach(fetchMock.restore)`. Doing it after tests is better. I think doing it before is a throwback to when restore was less error tolerant

winterstefan:
Finally, removed some unnecessary guard clauses and all `fetchMock.restore()` calls (where possible).

If there's anything to be done, don't hesitate contacting me.
Otherwise feel free to _squash&merge_ it. :-)

wheresrhys:
great effort - thanks for all this. Will release it as a patch

wheresrhys:
https://github.com/wheresrhys/fetch-mock/releases/tag/v5.10.1

PR #212, opened 2017-08-08
support combining function matchers with other options

### LINKED ISSUE #211: Unclear that a function matcher doesn't get to use method or limit options
We wrote a test like this:

```js
const allowedPath = matchUrl('/jobs');

fetchMock.deleteOnce(allowedPath, {});
fetchMock.getOnce(allowedPath, { jobs: [], availableRefinements: {}, total: 0 });
```

However, we kept getting `{}` returned from our GET request. After deep debugging, we realized that the function matcher was being called without regard to the HTTP method. Stepping up a level, we found [this comment](https://github.com/wheresrhys/fetch-mock/blob/01b9fc4333bade1363c49056da2b22b8296768d9/src/compile-route.js#L101-L102):

	// If user has provided a function as a matcher we assume they are handling all the
	// matching logic they need

From the outside, this isn't very clear at the usage sites of `fetchMock.get` or `getOnce`. The documentation says:

> `get()`
> Shorthands for `mock()` restricted to a particular method 

> `getOnce`
> Shorthands for `mock()` restricted to a particular method and that can only be called one time only

> `mock`
> - `matcher` Function(url, opts): A function (returning a `Boolean`) that is passed the url and opts `fetch()` is called with
> - `options`
> - - `method`: http method to match

None of these mention that a function style matcher will ignore the options passed to `mock`.

Additionally, the extra options (HTTP method, times, etc.) are not passed to the custom matcher to even allow us to implement that functionality ourselves.

<details>
<summary>The code for <code>matchUrl</code></summary>

Matches the actual URL when the path completely matches and all the specified query parameters match. The actual URL may have additional query parameters, they will be ignored.

```js
const matchUrl = expectedUrlStr => actualUrlStr => {
  const parseOne = u => url.parse(u, true);

  const expectedUrl = parseOne(expectedUrlStr);
  const actualUrl = parseOne(actualUrlStr);

  if (expectedUrl.pathname !== actualUrl.pathname) {
    return false;
  }

  for (const key in expectedUrl.query) {
    if (expectedUrl.query[key] !== actualUrl.query[key]) {
      return false;
    }
  }

  return true;
};
```

</details>

wheresrhys:
fixes #211 @shepmaster

### DIFF
--- a/src/compile-route.js
+++ b/src/compile-route.js
@@ -98,11 +98,7 @@ module.exports = function (route, Request, HeadersConstructor) {
 		route.__unnamed = true;
 	}
 
-	// If user has provided a function as a matcher we assume they are handling all the
-	// matching logic they need
-	if (typeof route.matcher === 'function') {
-		return route;
-	}
+	let matchUrl;
 
 	const expectedMethod = route.method && route.method.toLowerCase();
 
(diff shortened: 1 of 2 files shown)


shepmaster:
Thanks!

PR #222, opened 2017-10-04
Added includeContentLength to the configuration

Halt001:
I've added the `includeContentLength` option to both the global and the per route settings as you suggested. I've also added a couple of unit-tests and added a line to the documentation.

I hope this is what you had in mind.

P.S. One or two browser tests fail, but they also failed before I touched the code. This happens both on my local machine and on a Travis build.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -121,7 +121,8 @@ Returns the options for the last matched call to fetch
 
 #### `configure(opts)`
 Set some global config options, which include
-* `sendAsJson` [default `true`] - by default fetchMock will convert objects to JSON before sending. This is overrideable fro each call but for some scenarios e.g. when dealing with a lot of array buffers, it can be useful to default to `false`
+* `sendAsJson` [default `true`] - by default fetchMock will convert objects to JSON before sending. This is overrideable from each call but for some scenarios e.g. when dealing with a lot of array buffers, it can be useful to default to `false`
+* `includeContentLength` [default `false`]: When set to true this will make fetchMock automatically add the `content-length` header. This is especially useful when combined with `sendAsJson` because then fetchMock does the conversion to JSON for you and knows the resulting length so you don’t have to compute this yourself by basically doing the same conversion to JSON.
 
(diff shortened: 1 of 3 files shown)


test/spec.js
```diff
@@ -1259,6 +1313,61 @@ module.exports = (fetchMock, theGlobal, Request, Response) => {
...
+					includeContentLength: true
+				});
+				fetch('http://it.at.there/')
+					.then(res => {
+						expect(res.headers.get('content-length')).to.equal('100');
+						done();
+					});
+			});
+
+			it('should ignore case of explicit content-length header', done => {
```
wheresrhys: confusing title for the test - I read it as case meaning 'in the case that'. How about `should be case-insensitive when checking for explicit content-length header`

docs/api.md
```diff
@@ -36,6 +36,7 @@ Replaces `fetch()` with a stub which records its calls, grouped by route, and op
         * `headers`: Set the response headers. (`object`)
         * `throws`: If this property is present then a `Promise` rejected with the value of `throws` is returned
         * `sendAsJson`: This property determines whether or not the request body should be JSON.stringified before being sent (defaults to true).
+        * `includeContentLength`: Set this property to true to automatically add the `content-length` header (defaults to false).
```
wheresrhys: needs documenting under the `configure` method too

wheresrhys:
closing in favour of #223

wheresrhys:
@Halt001 I'm working on v6 of fetch-mock. Interested to hear your thoughts on whether content-length should default to always being added?

Halt001:
The `sendAsJson` option is already on by default so I think it would make sense to have `includeContentLength` also default enabled by default. The only reason I didn't do that is because it would change current behaviour. So it probably depends on how much behaviour you want to change in the new version. It is probably no big deal to go back to the old behaviour if needed because there is a global setting, so my initial thought is to indeed make that change.

wheresrhys:
yep, I was leaning the same way

Issue #227, opened 2017-11-06
Match url with query params event if the params are not in the same order


shmilky:
Hi,
Would love it if the same matcher will handle both ```http://test.url?a=2&b=4``` and ```http://test.url?b=4&a=2```

wheresrhys:
This has been requested before. I'm a bit reluctant to implement it. If you can suggest an API for this we can agree on and submit a PR then I will consider adding it.

wheresrhys:
Have decided on the following API.

```js
fetchMock.mock('/url/without/qs', 200, {
  query: {
    field: 'value'
  }
});
```
This has the advantage of being compatible with any type of matcher

wheresrhys:
Done in v6

rajatvijay:
I see this was closed way back but seems like error logging can be improved. If the matching fails just because of the query params, some hint in the error logs can save a lot of time for devs

wheresrhys:
I have a WIP fro implementing debug logging. It's not simple though because the same code is used for both matching fetch calls and for retrieving those calls later when inspecting - implementing debug logging in a way that makes it clear which of those phases the log is occurring during will require a bit of a refactor. I'm not sure exactly which approach to take yet

PR #229, opened 2017-11-10
Match urls with query params event if the params are not in the same order

shmilky:
See https://github.com/wheresrhys/fetch-mock/issues/227

* For cases where the url is a valid one we'll do a less basic equality check by separating the query params and compare them separately

* Added basic test

* Add query-string package to parse the query strings.

### DIFF
--- a/src/compile-route.js
+++ b/src/compile-route.js
@@ -1,6 +1,7 @@
 'use strict';
 const glob = require('glob-to-regexp')
 const express = require('path-to-regexp');
+const queryParser = require('query-string');
 
 const stringMatchers = {
 	begin: targetString => {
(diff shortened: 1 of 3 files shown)


src/compile-route.js
```diff
@@ -62,6 +63,55 @@ function areHeadersEqual (currentHeader, expectedHeader) {
     return true;
 }
 
+const pathPattern = /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/;
+const queryPattern = /^(\?)?([0-9A-Za-z]+=[0-9A-Za-z]+(\&)?)+$/;
+
+function getUrlSections (url) {
```
wheresrhys: can't you use `URL.parse()` here https://nodejs.org/api/url.html

src/compile-route.js
```diff
@@ -62,6 +63,55 @@ function areHeadersEqual (currentHeader, expectedHeader) {
...
+    const pathPart = splitUrl[0];
+    const queryPart = splitUrl.length > 1 ? splitUrl.slice(1).join('?') : '';
+
+    return {
+        path: pathPart,
+        query: queryPart
+    }
+}
+
+function isValidUrl (checkedUrl) {
```
wheresrhys: Similar to above, if you use the URL class, there will probably be no need for this custom validation

src/compile-route.js
vitalets: Maybe consider url normalization package? e.g. https://github.com/sindresorhus/normalize-url

wheresrhys:
I'm writing v6 of the library. I intend to support query string reordering via a similar mechanism to specifying headers i.e. if query order is unimportant, they should not be added in the string that matches the url, but rather passed in as a map in the options. This feels less magic and it's far clearer how this behaviour would blend with the various different methods of matching a url e.g. the basic url matcher could be a regex, but still support unordered query string matching.

Thanks for raising this and for the initial PR though

PR #234, opened 2017-12-10
Override Matcher Feature

ianwcarlson:
See #234 

Why:
* To provide the ability to override the response for a given matcher

How:
* Added logic to check for duplicate matchers when adding them. If a
  duplicate is found, then delete the old one and add the new one.
* Added unit tests to verify overriding behavior for string, regex,
  and function matchers.
* Removed the test for falling-back to previous matcher because it
  directly conflicts with the overriding feature.
* Modified the route name creation to include the method so that
  GET and POST mocks don't register as the same matcher
* Refactored compile-route.js to provide two named exports

### DIFF
--- a/src/fetch-mock.js
+++ b/src/fetch-mock.js
@@ -1,8 +1,12 @@
 'use strict';
 
-const compileRoute = require('./compile-route');
+const compileRouteModule = require('./compile-route');
+const compileRoute = compileRouteModule.compileRoute;
+const buildMatcherName = compileRouteModule.buildMatcherName;
 
 const FetchMock = {};
+
+const NOT_FOUND = -1;
 
 FetchMock.config = {
 	includeContentLength: false,
@@ -160,6 +164,16 @@ FetchMock.addRoute = function (route) {
 	if (!route) {
 		throw new Error('.mock() must be passed configuration for a route')
 	}
+
+  	const foundIndex = this.routes.findIndex(registeredRoute => {
+    		return buildMatcherName(route) === registeredRoute.name
+  	});
+
+  	// If duplicate matcher found, delete the old one. This provides the ability to
+  	// override the response for a specific matcher
+  	if (foundIndex !== NOT_FOUND) {
+    		this.routes.splice(foundIndex, 1);
+  	}
 
 	// Allows selective application of some of the preregistered routes
 	this.routes.push(compileRoute(route, this.config.Request, this.config.Headers));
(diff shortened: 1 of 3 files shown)


test/spec.js
```diff
@@ -1075,21 +1075,40 @@ module.exports = (fetchMock, theGlobal, Request, Response) => {
 						})
 				});
 
-				it('falls back to second route if first route already matched enough times', function () {
```
ianwcarlson: I removed this test because falling-back now conflicts with the overriding behavior.  There can be only one unique matcher that exists in the `this.routes` array.

src/fetch-mock.js
```diff
@@ -161,6 +164,16 @@ FetchMock.addRoute = function (route) {
 		throw new Error('.mock() must be passed configuration for a route')
 	}
 
+  	const foundIndex = this.routes.findIndex(registeredRoute => {
```
ianwcarlson: `findIndex` is in the ES6 standard so it should transpile ok.

ianwcarlson:
Formatting the code was a little tricky.  Have you considered using Prettier?

ianwcarlson:
All the tests are passing but the linter config in the v6 branch needs some work

src/fetch-mock.js
wheresrhys: is this supported in node 4? I avoid using a lot of ES201* features in order to avoid the need to transpile for node

test/spec.js
wheresrhys: ah, I didn't realise I explicitly supported that behaviour - the API has grown so much over the years 😵 . It makes me think that the correct way to implement this is to accept an option `{override: true}` when adding the mock

wheresrhys:
Also, if you fancy adding prettier in this PR please do. Happy to give it a go

test/spec.js
ianwcarlson: I'm fine with that. I should be able to get to this over the holidays.

src/fetch-mock.js
ianwcarlson: looks like Travis runs with Node 4 so I think we're good.

ianwcarlson:
> Also, if you fancy adding prettier in this PR please do. Happy to give it a go

This should probably be done in a separate PR when there are no significant PRs outstanding. 
 Otherwise, resolving the conflicts can be a nightmare.

wheresrhys:
I've merged this and changed the behaviour to depend on an `overwriteRoutes` option.

I've removed the addition of method to route names. I may add a second parameter to all the call inspection methods to allow for filtering calls by headers and methods, but I don't think doing this via matcher name is the way to go; i'd like to preserve the simplicity that it's just the route's matcher .toString()

Issue #239, opened 2018-01-05
Support all fetch init options as matcher options


wheresrhys:
Currently `headers` and `method` are supported. It makes sense to add all the following:

- [ ] destination
- [x] referrer
- [x] referrerPolicy
- [ ] mode
- [x] credentials
- [ ] cache
- [ ] redirect
- [ ] integrity
- [ ] keepalive
- [ ] signal
  
Pull requests most welcome 😄

ianwalter:
So was signal added in https://github.com/wheresrhys/fetch-mock/pull/381?

Issue #238, opened 2018-01-04
Configure the mock to match the credentials option.


squeedee:
currently I construct my mock like this:
```javascript
fetchMock.getOnce(
                "/api/user",
                {
                    body: {name: "jbloggs"},
                    headers: {"content-type": "application/json"},
                }
            )
```

and test the call to ensure the 'same-origin' credential option was passed:

```javascript
expect(fetchMock.lastOptions("/api/user").credentials).toEqual("same-origin")
```

I would prefer not to need the expect:

```javascript
fetchMock.getOnce(
                "/api/user",
                {
                    body: {name: "jbloggs"},
                    headers: {"content-type": "application/json"},
                },
                { credentials: "same-origin" }      
            )
```

wheresrhys:
You can use a function matcher, so

```js
fetchMock.mock((url, opts) => url === '/api/user' && opts.credentials === 'same-origin', 200)
```
The above is better than what you have now, but I like the feature request, and will add it to v6 of the library (which is in beta at the moment)

wheresrhys:
Have created a todo list with this on it https://github.com/wheresrhys/fetch-mock/issues/239

squeedee:
Thanks!

PR #241, opened 2018-01-07
Add support for matching referrer, referrerPolicy and credentials options

leggsimon:
Handles some of #239 (Also #238)
  

Just to double check the case here, am I right in thinking this should pass?
```js
it('match when credentials option isn\'t passed as a matcher', async () => {
	fm
		.mock('http://it.at.there/', 200)
		.catch();

	await fm.fetchHandler('http://it.at.there/', { credentials: 'same-origin' })
});
```

Where the mock isn't passed an option to match on `credentials` but it is given to `fetch` as an option.

### DIFF
--- a/test/specs/routing.test.js
+++ b/test/specs/routing.test.js
@@ -402,6 +402,65 @@ module.exports = (fetchMock) => {
 					});
 				})
 			});
+
+			describe('referrer', () => {
+				it('not match when referrer not present', async () => {
+					fm
+						.mock('http://it.at.there/', 200, {
+							referrer: 'http://referrer.example.com'
+						})
+						.catch();
+
+					await fm.fetchHandler('http://it.at.there/')
+					expect(fm.calls(true).length).to.equal(0);
+				});
+
+				it('not match when referrer does\'t match', async () => {
+					fm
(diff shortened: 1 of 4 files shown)


test/specs/routing.test.js
```diff
@@ -402,6 +402,210 @@ module.exports = (fetchMock) => {
...
+						.mock('http://it.at.there/', 200, {
+							referrer: 'http://referrer.example.com'
+						})
+						.catch();
+
+					await fm.fetchHandler('http://it.at.there/', { referrer: 'http://referrer.example.com' })
+					expect(fm.calls(true).length).to.equal(1);
+				});
+
+				it('be case insensitive', async () => {
```
wheresrhys: should it be case insensitive? I guess probably, but worth checking if you haven't already

leggsimon:
RE CI, looks like I messed up trying to fix the dependencies issue in bitHound, I'll revert that commit

wheresrhys:
d'you want to revert the dep changes then I'll merge the PR

wheresrhys:
bingo! This fixes your feature request @squeedee. I won't be able to release until this weekend though

PR #247, opened 2018-01-16
rename overrideRoutes to overwriteRoutes

olarclara:
The output message when there are equal routes is `Adding route with same name as existing route`. It might be a little confusing as the option is `overwriteRoutes`.


wheresrhys:
thanks for the fix

PR #259, opened 2018-02-01
update to check for a url object rather than a string

Jmeyering:


Jmeyering:
Was running into some issues with getting testing to run locally, so I didn't write any tests yet, but wanted to present this ahead of time to see if this was something to consider. One issue I thought of was that this is potentially a major BC breaking change (if people were relying on the unusual `url.url` functionality in their code.)

wheresrhys:
Thanks for submitting this fix. A test would be great too when you're able.

v6 of the library is only a couple of weeks old so I doubt there will be [m]any people relying on this broken behaviour, so definitely gonna treat it as a patch release when fixed.

LKay:
Is there anything stopping this from being merged and released? Calling `fetch(new Request(...))` is one of the most common use cases and right now it's totally broken in `fetch-mock` to handle this.

Jmeyering:
I believe tests need to be written for my fix. I've been unable to get to
it so far. If you feel able feel free to take a crack at it.

wheresrhys:
closing in favour of https://github.com/wheresrhys/fetch-mock/pull/278. Thanks for starting this @Jmeyering , and apologies for the delay merging

PR #264, opened 2018-02-19
check URL when fetch uses Request object instead string

alair-tavares-avalara:
Tests fail when to pass Request object as param in fetch.

`fetchMock.getOnce(API_URL, { body });`

>> Unmatched GET to [object Object]

`const request = new Request(API_URL, {
            method: 'GET',
            headers: headers,
            body: JSON.stringify(body)
        });`
`fetch(request).then(response => response.json()).catch(error => error);`

### DIFF
--- a/src/lib/compile-route.js
+++ b/src/lib/compile-route.js
@@ -3,24 +3,30 @@ const express = require('path-to-regexp');
 const URL = require('url');
 const querystring = require('querystring');
 
-function normalizeRequest (url, options, Request) {
-	if (Request.prototype.isPrototypeOf(url)) {
+function normalizeRequest(input, options, Request) {
+	if (checkRequestInstance(input, Request)) {
+		const { url, method, headers } = input;
 		return {
-			url: url.url,
-			method: url.method,
+			url,
+			method,
 			headers: (() => {
-				const headers = {};
-				url.headers.forEach(name => headers[name] = url.headers.name);
-				return headers;
+				const mappedHeaders = {};
+				headers.forEach(name => mappedHeaders[name] = headers.name);
+				return mappedHeaders;
 			})()
 		};
 	} else {
 		return {
-			url: url,
+			url: input,
 			method: options && options.method || 'GET',
 			headers: options && options.headers
 		};
 	}
+}
+
+function checkRequestInstance(url, configRequest) {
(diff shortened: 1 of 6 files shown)


src/lib/fetch-handler.js
```diff
@@ -3,6 +3,7 @@ const ResponseBuilder = require('./response-builder');
 const FetchMock = {};
 
 FetchMock.fetchHandler = function (url, opts) {
+	url = url instanceof Request ? url.url : url;
```
wheresrhys: rather than `Request` use `this.config.Request` as this is resilient to people using custom `fetch` implementations

src/lib/response-builder.js
```diff
@@ -10,7 +10,7 @@ const responseConfigProps = [
 
 module.exports = class ResponseBuilder {
 	constructor (url, responseConfig, fetchMock) {
-		this.url = url;
+		this.url = url instanceof Request ? url.url : url;
```
wheresrhys: rather than `Request` use `fetchMock.config.Request` as this is resilient to people using custom `fetch` implementations

wheresrhys:
Thanks for submitting. In addition to addressing my comments please also include a test in this PR - all bug fixes should include tests to describe the bug and to make sure it doesn't break in future. Thanks

alair-tavares-avalara:
@wheresrhys, I solved without changes with following code:
´
beforeAll(() => {
        fetchMock.config.Request = Request;
    });
´

Feel free if you want to cancel this PR.

src/lib/compile-route.js
```diff
@@ -3,26 +3,32 @@ const express = require('path-to-regexp');
...
 		return {
-			url: url,
+			url: input,
 			method: options && options.method || 'GET',
 			headers: options && options.headers
 		};
 	}
 }
 
+function checkRequestInstance (url, configRequest) {
```
wheresrhys: could you rename this function to `isRequestInstance` and also move it to above the function it's called from

wheresrhys:
nudge nudge 😄 @alair-tavares-avalara

wheresrhys:
The more I look at this, I don't think it's a bug. 

In your test you don't assign your dummy `Request` to `fm.config`, so it fails. If you do assign to config, which is the intended usage, then the test passes

PR #275, opened 2018-03-13
Move node-fetch to dependencies

jpmonette:
`node-fetch` should be in the `dependencies`. If not, the module will return:

```sh
  ● Test suite failed to run

    Cannot find module 'node-fetch' from 'server.js'

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:169:17)
      at Object.<anonymous> (node_modules/fetch-mock/src/server.js:2:15)
```

Workaround is to include it in the parent project, which is not optimal imho.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "homepage": "http://www.wheresrhys.co.uk/fetch-mock",
   "dependencies": {
     "glob-to-regexp": "^0.3.0",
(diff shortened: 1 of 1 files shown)


wheresrhys:
This is a deliberate decision, and allows the parent project to control which version of node-fetch they install. Why would it be suboptimal to ensure the testing library uses the same node-fetch that the parent project uses?

Issue #261, opened 2018-02-07
Cannot find module 'node-fetch' from 'server.js'


jepiqueau:
package.json should be upgraded as node-fetch@2.0.0 has been released few days ago and it seems that node-fetch-alpha.9 does not exist anymore

wheresrhys:
thanks for the heads up. Updated now

biswa8998:
node-fetch is not getting installed while running npm install for fetch-mock. Is it supposed to be added in dependencies instead of devDependencies?

PR #294, opened 2018-04-05
Move node-fetch to dependencies

wkillerud:
Hi! Thanks a bunch for the work you've put into `fetch-mock`!

I ran into an issue related to #261 where `fetch-mock` would throw an error due to missing `node-fetch` since I didn't have `node-fetch` in my own `package.json`.

This commit moves `node-fetch` from `devDependencies` to `dependencies` so we are sure consumers of `fetch-mock` gets the required version installed.

Naturally a workaround is to just declare `node-fetch` as a direct dependency in my project, but it would be nice to avoid since I don't use it directly.


wheresrhys:
This is the second time someone' raised this exact PR - I should probably document that it's a deliberate decision not to include it. 

fetch-mock works with any version of node-fetch, so it's deliberately left up to the consumer to install. I may add it as a peer dependency (if it's possible to do this as well as have it as a dev dependency)

PR #313, opened 2018-05-16
V7

wheresrhys:


wheresrhys:
@rubenv I happened to notice you're working on updating the typescript definition for fetch-mock.

I will probably release v7 fairly soon (to fix some annoying edge case regressions that crept in in v6). The changes are not going to be too drastic, other than that responses are wrapped in ES6 Proxies - I'm not sure if this has an impact on typescripts ability to recognise the type? Would be great if you could test it out and I could make the proxy opt-in (it's not needed by most features of the library)

rubenv:
I'm certainly no expert, but I think Typescript does all of its work
statically, at compile time, so I don't think it'll make much of a
difference.

Thanks for your great library!

lourenci:
@wheresrhys is there some prediction to merge this? Thank you.

wheresrhys:
There are other minor breaking changes I want to include in v7 - haven't had time to implement yet. Won't be soon, but there is a pre-release of v7 you can use https://github.com/wheresrhys/fetch-mock/releases/tag/v7.0.0-alpha.1

PR #318, opened 2018-06-12
Fixed fallbackToNetwork returning broken promise

dennybaa:
Please, have a look! At least it works for me...
The problem is that even if generateResponse gets native fetch (async function) it wraps it again into a malfunctioning promise by ResponseBuilder.

This doesn't work as expected:

```js
fetchMock.config.fallbackToNetwork = true

let res = await fetch('blabla')

// print a mangled response,
// but it has to directly pass the native fetch response back!
console.log(await res.json())

```

### DIFF
--- a/src/lib/fetch-handler.js
+++ b/src/lib/fetch-handler.js
@@ -59,16 +59,14 @@ FetchMock.generateResponse = async function (response, url, opts) {
 	// - function returning a Promise for a response
 	// - delaying (using a timeout Promise) a function's execution to generate
 	//   a response
-	// Because of this we can't safely check for function before Promisey-ness,
-	// or vice versa. So to keep it DRY, and flexible, we keep trying until we
+	// Keep it DRY, and flexible, we keep trying until we
 	// have something that looks like neither Promise nor function
+
 	while (typeof response === 'function' || typeof response.then === 'function') {
-		if (typeof response === 'function') {
-			response = response(url, opts);
-		} else {
-			// Strange .then is to cope with non ES Promises... god knows why it works
-			response = await response.then(it => it)
-		}
+		// got promise or async function, since we don't need to build a response
+		// with simply return it (ex. in case fallbackToNetwork == true)
+		if (response instanceof Promise) return response;
+		else response = response(url, opts);
 	}
 
 	// If the response is a pre-made Response, respond with it


wheresrhys:
Sorry, I don't quite understand the problem you're describing. Could you add a failing test please.

As there are many `Promise` implementations available (e.g. Bluebird) it's also in general not safe to check for `instanceof Promise` because the promises used by the user may e.g. be different to the ones used by a fetch-based library they are using. I thought I had a test to cover this, so I'm surprised your code passes

wheresrhys:
Thanks for the additional details. I'll take a closer look later. After the football

wheresrhys:
wow - I have no tests that verify fallbackToNetwork works, only that when it is configured to be on, fetch-mock does not explicitly throw

Thanks for raising this issue

wheresrhys:
I'm struggling to write a failing test. This one is an attempt to capture the issue you describe, but it always passes https://github.com/wheresrhys/fetch-mock/compare/fallback-to-network

Could you provide some modifications that'll make it fail?

dennybaa:
@wheresrhys It can't pass IMO, at least with ES6 promises. You should write the test somehow or make a real non-mocked call. You could actually start some simple router directly from your tests, like [this](https://github.com/pillarjs/router).

```js
var finalhandler = require('finalhandler')
var http         = require('http')
var Router       = require('router')

var router = Router()
router.get('/', function (req, res) {
  res.setHeader('Content-Type', 'application/json; charset=utf-8')
  res.end('{"hello": "world"}')
})

var server = http.createServer(function(req, res) {
  router(req, res, finalhandler(req, res))
})

server.listen(3000)
```

And a sample test:

```js
...
    it('fallback to network must return unmodified response', async() => {
      fetchMock.config.fallbackToNetwork = true

      let res = await fetch('//localhost:3000/')
      res = await res.json()
      fetchMock.config.fallbackToNetwork = false

      expect(res).not.be.empty
      expect(res.hello).to.equal('world')
    })
```

Or any other way, these are just samples...

wheresrhys:
I've rewritten the test using a real server (I think I'll rewrite the whole test suite to use one, in fact), but still can't reproduce the bug.

Could you please create a repo I can clone and install to recreate the bug for myself. Thanks

wheresrhys:
Cheers 😄 

I'll close this PR for now in that case. If you discover anything worth sharing please consider submitting a PR for the troubleshooting section of the docs

PR #321, opened 2018-06-28
Add failing test when relative URLs are requested

adam-nielsen:
I've recently run into an issue where I'm unable to mock certain URLs which work fine in a web browser, as fetch-mock doesn't seem to support requests to relative URLs.

This PR is a failing test case which demonstrates the issue I am coming across.  Hopefully this explains the issue.

### DIFF
--- a/src/lib/compile-route.js
+++ b/src/lib/compile-route.js
@@ -1,12 +1,12 @@
 const glob = require('glob-to-regexp')
 const express = require('path-to-regexp');
-const URL = require('url');
+const nodeURL = require('url');
 const querystring = require('querystring');
 
 function normalizeRequest (url, options, Request) {
 	if (Request.prototype.isPrototypeOf(url)) {
 		return {
-			url: url.url,
+			url: normalizeURL(url.url),
 			method: url.method,
 			headers: (() => {
 				const headers = {};
@@ -16,10 +16,22 @@ function normalizeRequest (url, options, Request) {
 		};
 	} else {
 		return {
-			url: url,
+			url: normalizeURL(url),
 			method: options && options.method || 'GET',
 			headers: options && options.headers
 		};
+	}
+}
+
+function normalizeURL (url) {
+	var r = new RegExp('^(?:[a-z]+:)?//', 'i'); // https://stackoverflow.com/a/19709846/308237
+	const absolute = r.test(url);
+	if (absolute) {
+		const u = new URL(url);
+		return u.href;
+	} else {
+		const u = new URL(url, 'http://dummy');
+		return u.pathname;
 	}
 }
 
@@ -95,7 +107,7 @@ const getQueryStringMatcher = route => {
 	}
 	const keys = Object.keys(route.query);
 	return ({ url }) => {
-		const query = querystring.parse(URL.parse(url).query);
+		const query = querystring.parse(nodeURL.parse(url).query);
 		return keys.every(key => query[key] === route.query[key]);
 	}
 }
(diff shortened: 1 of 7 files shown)


wheresrhys:
So your issue is that you would expect `/it.at/not/../there/`to resolve to `/it.at/there/` and be successfully mocked? 

As far as I know, that sort of behaviour is dependent on the server running the site, and has nothing to do with specced http/fetch behaviour, so is outside the scope of this library. When not mocking, does your site respond with a 301/302 redirecting `/it.at/not/../there/` to `/it.at/there/`?

wheresrhys:
Well, I never knew that. So, yep, this looks like a valid bug. I'm not gonna have time to work on a fix at the moment though, so please add a solution to this PR if it's urgent. I would ask, though, that you keep the existing test you've edited unaltered and add a new test for removing dot segments.

adam-nielsen:
Here's an updated PR that is one possible solution, using the built in [URL](https://url.spec.whatwg.org/) library, so no additional dependencies should be needed.  The modified test is moved into a separate, additional test.

You're probably not going to like it though, because a side effect of this is that all URLs are normalised/canonicalised, which results in all URLs of the type `http://example.com` getting a trailing slash added.  So any existing `fetch-mock` code that is only mocking `http://example.com` needs to be updated to mock `http://example.com/` instead.  It will, however, fix #324.

This PR updates all the tests like this so that they pass, except for one test which still fails and I am not sure why.

I think that in some ways this is more "correct" as the URLs being mocked will always be explicit about the path (even if it's just `/`) however the drawback is potentially breaking everyone's existing tests until they update their mocked URLs.

As far as I am aware, when you request a URL with no path component, the browser requests the root path anyway (since it ends up as a `GET /` request) so adding the trailing slash wouldn't change the logic behind the URL.

src/lib/compile-route.js
```diff
@@ -16,13 +16,25 @@ function normalizeRequest (url, options, Request) {
...
 		return {
-			url: url,
+			url: normalizeURL(url),
 			method: options && options.method || 'GET',
 			headers: options && options.headers
 		};
 	}
 }
 
+function normalizeURL (url) {
```
wheresrhys: could you move this above where it's called please.

src/lib/compile-route.js
```diff
@@ -16,13 +16,25 @@ function normalizeRequest (url, options, Request) {
...
+	var r = new RegExp('^(?:[a-z]+:)?//', 'i'); // https://stackoverflow.com/a/19709846/308237
```
wheresrhys: could you move this comment above the line and add some text explaining it is testing for the part of the url before path

test/specs/inspecting.test.js
```diff
@@ -139,7 +139,7 @@ module.exports = (fetchMock) => {
 						name: 'my-route'
 					});
 
-				await fm.fetchHandler('http://it.at.here/');
+				await fm.fetchHandler('http://it.at.here');
```
wheresrhys: I feel it'd be better to keep ending slashes consistent within a single test, and only vary them in tests that explicitly test the trailing slash behaviour

test/specs/routing.test.js
```diff
@@ -533,6 +533,17 @@ module.exports = (fetchMock) => {
 				expect(fm.calls(true).length).to.equal(1);
 			});
 
+			it('match relative urls with dots', async () => {
```
wheresrhys: given that the src branches depending on whether it's an absolute or relative url, I'd like to see a test for each case

wheresrhys:
Closing in favour of https://github.com/wheresrhys/fetch-mock/pull/332, which does the tricky merge with prettified code.
Thanks for your contribution

Issue #322, opened 2018-06-28
Global switch for dev mode


janukobytsch:
Hey! Is there a way to disable all mocks in a single call? I had a look at the global options, but couldn't find anything. Basically I need an option like `fallbackToNetwork` that always falls back irregardless of whether there is a matcher defined. The use case for this is migrating from local stubs to the real API.

wheresrhys:
There isn't i'm afraid. If you want a hacky way to do so you could set 
```javascript
fetchMock.config.fallbackToNetwork = true;
fetchMock.addRoute = function () {
   this.routes = [];
}
```
... that _should_ do the job.

A PR to do this properly would be most welcome. Something like accepting `always` as a value for `fallbackToNetwork`, and this skips routing entirely

dennybaa:
Maybe you could smth like this, this just an idea...

```js
const mockedInstance = fetchMock.createInstance()

// enable mocks function
mockedInstance.enable = () => {
    // mockedInstance.mock(...)
    // mockedInstance.mock(...)
}

mockedInstance.config.fallbackToNetwork = true

export {mockedInstance as default}
```

janukobytsch:
Thanks! Monkeypatching `addRoute` on the instance is working fine for me. Since this is a bit hacky, I created a PR based on @wheresrhys suggestion.

PR #323, opened 2018-06-28
Add global switch to always fallback to network

janukobytsch:
Based on #322


wheresrhys:
Great - thanks for submitting. I will release this shortly as a minor release

PR #324, opened 2018-06-29
Add test for matching 'end:' with Request object

adam-nielsen:
This test exposes an issue with the `end:` matcher where it doesn't always work if the `fetch()` call has been passed a `Request` object instead of a string.

The problem is that the URL coming from the `Request` object is normalised, which results in it getting a slash appended to it.  This extra slash causes the string matcher (compile-route.js:31) to compare `here/` against `there` which is false, resulting in the `end:` matcher failing.

The same test but using strings instead of `Request` objects passes (as shown by the test directly before the new one in this PR, which I have copied to make the new test.)

### DIFF
--- a/test/specs/routing.test.js
+++ b/test/specs/routing.test.js
@@ -50,6 +50,18 @@ module.exports = (fetchMock) => {
 				expect(fm.calls(true).length).to.equal(1);
 			});
 
+			it('match end: keyword with Request', async () => {
+				fm
+					.mock('end:there', 200)
+					.catch();
+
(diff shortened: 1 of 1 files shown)


wheresrhys:
Hmmm, that's an annoying edge case. I can't see how it can be fixed as, when fetching from a host without a path, `Request` is a two to one mapping, so can't be deterministically reversed. I'll leave this open while deciding how/whether to document this. Thanks for exploring it

wheresrhys:
Closing as it's now implemented in the v7 branch

PR #328, opened 2018-07-10
Added support for signal option

dutradda:
Implements the signal checkpoint of the issue https://github.com/wheresrhys/fetch-mock/issues/239

### DIFF
--- a/src/lib/fetch-handler.js
+++ b/src/lib/fetch-handler.js
@@ -21,6 +21,12 @@ FetchMock.fetchHandler = function (url, opts) {
 	// wrapped in this promise to make sure we respect custom Promise
 	// constructors defined by the user
 	return new this.config.Promise((res, rej) => {
+		if (opts && opts.signal) {
+			opts.signal.addEventListener('abort', () => {
+				rej(new Error(`URL '${url}' aborted.`));
+			});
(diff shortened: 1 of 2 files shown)


wheresrhys:
Thanks for implementing this. Wow - cancellable promises landed quickly.

Before merging I'd like to see some tests for how this impacts the `flush()` and `done()` methods. Maybe some other aspects of the API too. In particular, because generating a response can kick off a chain of microtasks, whereas `fetch` will - I assume - only ever create one, I wonder if there's potential to time the sending of the abort signal in such a way that it'd be too late to stop a genuine fetch, but would be able to intercept fetch-mock. A pretty edge case, but worth at least giving some thought to

wheresrhys:
Hi @dutradda. Did you see my response?

dutradda:
Hello @wheresrhys ,

thank's for the reply

I will do the tests with `flush()` and `done()` calls

thommyhh:
@dutradda @wheresrhys What is the status here? I could really use this right now.

wheresrhys:
@thommyhh I'm rewriting documentation for v7. Unless @dutradda does so first, I plan on completing this PR after that's done and it will be released in v7

PR #333, opened 2018-07-22
Move prettier to devDependencies

goto-bus-stop:
So users of `fetch-mock` don't install it. Thanks!


wheresrhys:
good spot. thanks

Issue #336, opened 2018-08-07
Requests that pass an URL object to fetch do not get matched


m90:
When wanting to perform a GET request that includes query parameters the [fetch standard recommends](https://fetch.spec.whatwg.org/#fetch-api) constructing a `new window.URL`, appending to its `searchParams` and then directly passing this URL object to `fetch`. From the doc:

```js
var url = new URL("https://geo.example.org/api"),
    params = {lat:35.696233, long:139.570431}
Object.keys(params).forEach(key => url.searchParams.append(key, params[key]))
fetch(url).then(/* … */)
```

I tried using `fetch-mock` to mock this kind of request, but found that it would not match the request until I explicitly called `.toString()` on the URL object. Passing an URL object to the matcher would not work either.

I.e. I have code like:

```js
exports.fetchURL = (url) => {
	const u = new URL(url);
	return fetch(u).then(res => res.json());
}
```

that will be tested llike this:

```js
describe('fetchURL', () => {
	context('with string matcher', () => {
		before(() => {
			fetchMock.get('http://api.example.com/resource/123', { ok: true });
		});
		after(() => {
			fetchMock.restore();
			fetchMock.reset();
		});
		// fails with Error: No fallback response defined for GET to http://api.example.com/resource/123
		it('resolves with the payload', () => {
			return fetchURL('http://api.example.com/resource/123').then((payload) => {
				assert.deepStrictEqual(payload, { ok: true });
			});
		});
	});

	context('with URL matcher', () => {
		before(() => {
			fetchMock.get(new URL('http://api.example.com/resource/123'), { ok: true });
		});
		after(() => {
			fetchMock.restore();
			fetchMock.reset();
		});
		// fails with TypeError: route.matcher.indexOf is not a function
		it('resolves with the payload', () => {
			return fetchURL('http://api.example.com/resource/123').then((payload) => {
				assert.deepStrictEqual(payload, { ok: true });
			});
		});
	});
});
```

but both tests fail. I'm not too familiar with the spec, but me and my guts would assume that both tests pass.

I created an example repo for the issue here: https://github.com/m90/fetch-mock-url

In case this really is a bug, I'd be happy to fix this myself in case I get one or two pointers to start from.

m90:
So it looks like `v7` already fixed the issue without knowing about it.

This test:

```js
it('match URL', async () => {
	fm.mock('http://it.at.there/', 200).catch();

[...]

PR #337, opened 2018-08-08
Add test for matching against URL object

m90:
See #336

### DIFF
--- a/test/specs/routing.test.js
+++ b/test/specs/routing.test.js
@@ -1,4 +1,5 @@
 const chai = require('chai');
+const URL = require('whatwg-url');
 const expect = chai.expect;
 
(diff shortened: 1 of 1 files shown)


test/specs/routing.test.js
```diff
@@ -291,14 +299,26 @@ module.exports = fetchMock => {
 
 			describe('query strings', () => {
 				it('can match a query string', async () => {
-					fm.mock('http://it.at.there/', 200, { query: { a: 'b' } }).catch();
```
m90: Prettier wouldn't let me write it like this anymore, so I decided to normalize the two.

wheresrhys: :+1:

PR #343, opened 2018-08-14
Added body to the normalized request

jww:
Added body to the normalized request in order to make it so that function matchers have the body contents available to be considered when routing


wheresrhys:
Thanks for this @jww. I'm confused why this test hasn't been failing https://github.com/wheresrhys/fetch-mock/blob/master/test/specs/routing.test.js#L109, so I think I need to spend some time understanding what's gone wrong before merging this PR

wheresrhys:
I don't think this is an issue any more. Let me know if that isn't the case

PR #376, opened 2018-10-02
375-fix-relative-url-with-query-bug

sepo-one:
Fixes:
- query matcher with relative url path
- route overwrite


sepo-one:
Similar goes for methods:
```
it('configure an exact method to match', async () => {
                fm.mock('http://it.at.there/', 200, { method: 'GET' })
                    .mock('http://it.at.there/', 200, { method: 'POST' })
                    .catch();

                await fm.fetchHandler('http://it.at.there/', { method: 'GET' });

                await fm.fetchHandler('http://it.at.there/', { method: 'POST' });
            });
```

@wheresrhys  am i missing anything ?

PR #378, opened 2018-10-03
Support names that include capitals

michaeldfallen:
This change allows names and identifiers to contain capitals, which is
fairly common behaviour for anyone who uses camelcase style when coding.


wheresrhys:
:wave: hello there. Thanks for adding this. Will release it shortly

michaeldfallen:
Awesome, thanks 🌟

PR #382, opened 2018-10-18
Remove node-fetch from peerDependencies since it is optional

### LINKED ISSUE #377: node-fetch in peerDependencies but we don't use it
Hi!

After updating to fetch-mock from v6 to v7, we now get a missing peer dependency warning when installing:

```
warning " > fetch-mock@7.0.6" has unmet peer dependency "node-fetch@*".
```

However we only use fetch-mock to mock native browser usage, so have no need to install the node-fetch package other than to silence the warning.

Would it be possible to:
* consider removing the peer dep, to prevent the warning for the browser mocking use-case
* or if that's not desired, then add a mention about it in the upgrading guide? (technically adding a peer dep is a breaking change, since it can make certain package managers exit non-zero when they didn't before)

Many thanks :-)

edmorley:
fetch-mock v7 added a peer dependency on `node-fetch`, however the package is not required for browser-based usages of fetch-mock. Such projects must then either install the dependency even if they don't use it, or else have peer dependency warnings be output by their package manager.

Hopefully in the future package managers will have a way to declare optional peer dependencies (eg yarnpkg/yarn#6487), so version ranges can be enforced iff people are using the dependency - but for now it's best to omit them from `peerDependencies` entirely.

Fixes #377.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,8 +81,5 @@
     "sinon-chai": "^2.14.0",
     "webpack": "^4.0.0",
     "webpack-cli": "^3.0.0"
-  },
-  "peerDependencies": {
-    "node-fetch": "*"
   }
 }


wheresrhys:
Thanks

PR #383, opened 2018-10-24
Docs: Added a use case (mocking a request object in a test) to the tr…

ReasonableDeveloper:
I was writing some Node.js unit tests for my service-workers and kept getting the `Error: Unrecognised Request object`. This was caused by me mocking the `Request` object. I searched for a solution and read through the troubleshooting page. When I read about using npm version < 3 I skipped the rest of the document that could have given me the solution to my problem and spared me another 10 minutes. I hope my update makes it more clear for others running into the same problem.

Note that the troubleshooting section for the legacy npm version could probably be skipped entirely if you want to change the supported node version to >=5 at some point. https://nodejs.org/en/download/releases/.
```
"engines": {
  "node": >=5
}
```


wheresrhys:
Thanks for this - I wish more people would improve the docs for me 😄

PR #400, opened 2018-12-07
tests to retrieve fetches to make sure header objects and header inst…

ReasonableDeveloper:
Hey, I created some tests because I ran into an issue described below. It turned out that in my mocked `headers` implementation, I was returning `Object.values()`  instead of `Object.entries()` in my `[Symbol.iterator]` function, causing some trouble on request normalization. Below is the issue for anyone running into the same problem. I was about to create the issue until I saw that the tests that were already there and the ones I added were passing and it was my implementation that was wrong. Feel free to merge the tests if you don't think they are already redundantly covered.

I'm running into a funny issue where my {key: value} header gets changed. The original key of my header is omitted while the first character of the value becomes the header key and the second character of the value becomes the new header value after the request gets normalized.
E.g if I were to pass `{'Authorization': 'JWT TOKEN'}` as a header object, it would be turned into `{ 'J': 'W' }` when retrieving the mocked fetch.
The problem seems to be the `headersToArray` function which is called before zipping the object. My mocked header object is an iterator; hence, the object gets spread into an array where the key is completely omitted. https://github.com/wheresrhys/fetch-mock/blob/dc7d5e2c7e58e3dd3f7610e08c6213044bc3eaca/src/lib/request-utils.js#L10
I'm not sure if there is something wrong with my mocked header implementation but if I'm not mistaken, [...headers]  will always simply return the values and no key, value pairs.
^ I was mistaken because I didn't understand iterables 😄

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3328,11 +3328,13 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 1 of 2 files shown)


wheresrhys:
Yep, I think filtering calls be header is tested lower down. Thanks for the PR though, and glad you fixed your problem

PR #405, opened 2018-12-24
Move karma to the dev dependencies

koenvg:
Moved karma to the dev dependencies


wheresrhys:
oops. Thanks for fixing

PR #414, opened 2019-02-14
Respect the loglevel for postinstall

stphnrdmr:
This adds support for disabling the postinstall message on CI and when the loglevel asks for it.


wheresrhys:
Great addition - thanks

PR #418, opened 2019-03-17
AbortError implementation

### LINKED ISSUE #416: Does not conform to the spec when aborting an error
When a request is aborted, `fetch-mock` throws a regular `Error`, but the spec (and `node-fetch` behaviour) is to throw an `AbortError`.

Spec: https://dom.spec.whatwg.org/#aborting-ongoing-activities
> APIs that rely upon AbortController are encouraged to respond to abort() by rejecting any unsettled promise with a new "AbortError" DOMException.

Code in node-fetch throwing `AbortError` (correct): https://github.com/bitinn/node-fetch/blob/master/src/index.js#L53
```
const abort = ()  => {
	let error = new AbortError('The user aborted a request.');
	reject(error);
	if (request.body && request.body instanceof Stream.Readable) {
		request.body.destroy(error);
	}
	if (!response || !response.body) return;
	response.body.emit('error', error);
}
```

Similar code in `fetch-mock` throwing an `Error` (incorrect): https://github.com/wheresrhys/fetch-mock/blob/master/src/lib/fetch-handler.js#L42

```
if (options && options.signal) {
	const abort = () => {
		rej(new Error(`URL '${url}' aborted.`));
		done();
	};
	if (options.signal.aborted) {
		abort();
	}
	options.signal.addEventListener('abort', abort);
}
```

--- 
Example code:

```
// WORKING, outputs 'Aborted'
const fetch = require('node-fetch');

// NOT WORKING, outputs 'Other error'
// const fetch = require('fetch-mock').sandbox();
// fetch.mock('*', 200);

const AbortController = require('abort-controller');
const controller = new AbortController();
fetch('http://www.google.com', { signal: controller.signal }).catch(err => {
    if (err.name === 'AbortError') console.log('Aborted!');
    else console.log('Other error');
});
controller.abort();
```

scinos:
Fixes #416

Returns an AbortError instead of a plain Error when aborting a request

### DIFF
--- a/src/lib/fetch-handler.js
+++ b/src/lib/fetch-handler.js
@@ -2,6 +2,16 @@ const responseBuilder = require('./response-builder');
 const requestUtils = require('./request-utils');
 const FetchMock = {};
 
+class AbortError extends Error {
+	constructor() {
+		super(...arguments);
+		this.name = 'AbortError';
+
+		// Do not include this class in the stacktrace
+		Error.captureStackTrace(this, this.constructor);
+	}
+}
+
 const resolve = async (response, url, options, request) => {
 	// We want to allow things like
 	// - function returning a Promise for a response
@@ -41,7 +51,7 @@ FetchMock.fetchHandler = function(url, options, request) {
 	return new this.config.Promise((res, rej) => {
 		if (options && options.signal) {
 			const abort = () => {
-				rej(new Error(`URL '${url}' aborted.`));
+				rej(new AbortError(`URL '${url}' aborted.`));
 				done();
 			};
 			if (options.signal.aborted) {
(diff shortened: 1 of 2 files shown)


src/lib/fetch-handler.js
```diff
@@ -2,6 +2,16 @@ const responseBuilder = require('./response-builder');
 const requestUtils = require('./request-utils');
 const FetchMock = {};
 
+class AbortError extends Error {
+	constructor() {
+		super(...arguments);
+		this.name = 'AbortError';
```
caub: I don't think you need that line, the name will already be 'AbortError'

temyers:
Is there any movement on this?

LaurensBosscher:
Any progress on this? I ran into this today :)

wheresrhys:
Closing in favour of https://github.com/wheresrhys/fetch-mock/pull/440

Sorry for taking so long to address this, and thanks for doing most of the work @scinos

PR #420, opened 2019-04-01
remove unnecessary ._calls, and package-lock not needed for libs

caub:


wheresrhys:
Thanks for tidying up after me 😄

PR #429, opened 2019-05-22
Use babel transform runtime over babel polyfill

evan-10e:
### Problem
Projects that depend on babel-polyfill and fetch-mock will crash with the following [error](https://github.com/babel/babel/issues/4019).

### Solution

**In short:** Libraries should not be using babel-polyfill, but rather babel transform runtime.

### Additional Context
Babel team members warn that if you are writing a library - babel polyfill should not be used because it can pollute the globals used by project owners.

> If you are writing a library, you should not be using babel-polyfill. Having a library mutate the global state of the browser environment is not a good idea. If you need a standalone library you should be using babel-runtime via transform-runtime, or if you don't want to do that, you should make it a requirement that uses of your library load a polyfill first, and your code should assume it is there.             
>      
> Logan Smyth (member and contributor of babel) https://github.com/babel/babel/issues/4019#issuecomment-245409644

As such, Fetch-mock's improper use of babel is causing projects to crash in this fashion. https://github.com/babel/babel/issues/4019#issuecomment-245409644

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1050,6 +1050,7 @@
       "version": "6.26.0",
       "resolved": "https://registry.npmjs.org/babel-polyfill/-/babel-polyfill-6.26.0.tgz",
       "integrity": "sha1-N5k3q8Z9eJWXCtxiHyhM2WbPIVM=",
+      "dev": true,
       "requires": {
         "babel-runtime": "^6.26.0",
         "core-js": "^2.5.0",
@@ -9693,7 +9694,8 @@
     "regenerator-runtime": {
       "version": "0.10.5",
       "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.10.5.tgz",
-      "integrity": "sha1-M2w+/BIgrc7dosn6tntaeVWjNlg="
+      "integrity": "sha1-M2w+/BIgrc7dosn6tntaeVWjNlg=",
+      "dev": true
     },
     "regenerator-transform": {
       "version": "0.10.1",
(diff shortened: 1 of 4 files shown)


wzph-10e:
+1

wheresrhys:
Hi. Sorry for the slow response. Not had much time to maintain the library lately.

I agree with your appraisal, but don't think I can release this as a patch - babel polyfilling has been a discussion point in issues before, and I think this will break lots of people's tests. I don't want to deal with the fallout from that.

So I think this should be released as v8 of the library. Please could you search the repo (mainly docs) for mentions of v7 and update to say v7/8 https://github.com/wheresrhys/fetch-mock/search?q=v7&unscoped_q=v7. I will add a changelog note once released to explain what the difference between v7 and 8 is.

ndelangen:
Could this be combined with an upgrade to core-js v3?

Or would this remove the need for core-js entirely?

Some users of storybook are hitting the core-js 3 incompatibility with this lib:
https://github.com/storybookjs/storybook/issues/7445

I would be open to helping resolve this, Are you still working on this @evan-10e @wheresrhys ?

evan-10e:
@ndelangen I discovered that my bundle script was duplicated in the page head. Removing the dup solved the issue for me.

wheresrhys:
[v8.0.0-alpha.1](https://www.npmjs.com/package/fetch-mock/v/8.0.0-alpha.1) attempts to address this issue. Please leave a :+1: on the pull request https://github.com/wheresrhys/fetch-mock/pull/457 if your problem is solved, or alternatively leave a comment @evan-10e @ndelangen @wzph-10e

PR #431, opened 2019-05-28
keeping arbitrary options on fetch

### LINKED ISSUE #430: Support checking for arbitrary options on .calls
I'm working on a Cloudflare Worker script that uses arbitrary options on a fetch call (as specified [here](https://developers.cloudflare.com/workers/reference/cloudflare-features/)).

However, I'm unable to test if those options are set correctly because fetch-mock removes them from the options list (as seen [here](https://github.com/wheresrhys/fetch-mock/blob/527d170ecf86b0054d85acfcc7e015b332e78f6f/src/lib/request-utils.js#L41)).

*Example:*

```javascript
function doFetchSomething(url) {
   fetch(url, { cf: { ttl: 30 } })
}

test('should add cf options', () => {
   const fetchMock = require('fetch-mock');
   const url = 'http://some.test.url.io';

   fetchMock.get(url, {});

   doFetchSomething(url);

   expect(fetchMock.calls(filter)[0][1]).toEqual({"cf": {"ttl": 30}, "headers": { "method": "GET"});
});

// Expected: {"cf": {"ttl": 30}, "headers": { "method": "GET"}
// Received: {"headers": { "method": "GET"}
```

nynkelys:
closes #430

### DIFF
--- a/src/lib/request-utils.js
+++ b/src/lib/request-utils.js
@@ -34,7 +34,7 @@ const normalizeUrl = url => {
 };
 
 module.exports = {
-	normalizeRequest: (url, options={}, Request) => {
+	normalizeRequest: (url, options, Request) => {
 		if (Request.prototype.isPrototypeOf(url)) {
 			const obj = {
 				url: normalizeUrl(url.url),
--- a/src/lib/request-utils.js
+++ b/src/lib/request-utils.js
@@ -34,13 +34,16 @@ const normalizeUrl = url => {
 };
 
 module.exports = {
-	normalizeRequest: (url, options, Request) => {
+	normalizeRequest: (url, options={}, Request) => {
 		if (Request.prototype.isPrototypeOf(url)) {
 			const obj = {
 				url: normalizeUrl(url.url),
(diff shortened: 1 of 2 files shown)


wheresrhys:
I was a _little_ nervous of this. Cloudflare are calling fetch with a `Request` instance and an `options` object, which I've never seen done before (if using a `Request` instance, you'd normally pass the options into the `Request` constructor), so I thought it might be a non-standard extension of the API. But have tried this scenario in Chrome, and it seems to work, blending the new options with those previously existing in the Request
image

... so I guess it's fine.

My only problem with this PR in that case is that the tests you've written are testing the wrong code path; your tests test passing a url with options, not a `Request` with options, and the library already supported (accidentally) arbitrary options in the former case. Could you add a new test where you call fetch using a Request?

Exciting it's been used to test cloudflare workers. Yet another environment where fetch is being used

wheresrhys:
Great thanks. Will release it shortly as a patch

nynkelys:
Thanks Rhys!

Issue #419, opened 2019-03-26
ES5 bundle conflict with core-js v3


robertknight:
The core-js project released a new version, [v3.0.0](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md), which changes the module layout. The generated fetch-mock es5 bundle in `es5/lib.js` requires core-js modules assuming core-js v2's layout. As a result, attempting to include fetch-mock in a browserify-generated bundle (eg. for use with a browser-based test runner such as Karma) fails.

**Steps to reproduce**

1. Save the following `package.json` into an empty directory:

```json
{
  "name": "fetch-mock-corejs-v3-test",
  "version": "1.0.0",
  "devDependencies": {
    "browserify": "^16.2.3",
    "core-js": "^3.0.0",
    "fetch-mock": "^7.3.1"
  }
}
```
2. Run `npm install`
3. Run `./node_modules/.bin/browserify -r fetch-mock`

**Output**

```
Error: Cannot find module 'core-js/modules/es6.typed.array-buffer' from '/private/tmp/fetch-mock-corejs-v3-test/node_modules/fetch-mock/es5'
    at /private/tmp/fetch-mock-corejs-v3-test/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:17
    at process (/private/tmp/fetch-mock-corejs-v3-test/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/private/tmp/fetch-mock-corejs-v3-test/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/private/tmp/fetch-mock-corejs-v3-test/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/private/tmp/fetch-mock-corejs-v3-test/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /private/tmp/fetch-mock-corejs-v3-test/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqCallback.oncomplete (fs.js:158:21)
```

**Notes**

My suggested solution here would be to omit polyfills from fetch-mock itself and instead make it the responsibility of the package consumer to include polyfills, at least for ES standard library functionality. This avoids potential conflicts with other polyfills and in some cases will reduce bundle size.

Thoughts? I would be happy to submit a PR if the proposed change is agreeable.

[...]

PR #432, opened 2019-06-11
Fixing dependent libs that need core-js@3 from breaking by adding a local dependency.

travist:
I know this seems counter intuitive, but by just adding the dependency of this library on core-js@2, you force NPM to install a local dependency "node_modules" folder which will use that version of core-js instead of using the dependents version.

This small change solved our dependency issue when we upgraded to core-js@3.

This fixes https://github.com/wheresrhys/fetch-mock/issues/419

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "homepage": "http://www.wheresrhys.co.uk/fetch-mock",
   "dependencies": {
     "babel-polyfill": "^6.26.0",
+    "core-js": "^2.6.9",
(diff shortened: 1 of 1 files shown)


zzak:
Curious if it's much work to just upgrade `fetch-mock` to support core-js@3?

Haven't looked into this myself, but have others?

wheresrhys:
OK - demand for this is quite high, so I'm gonna take a risk and merge without a major release. If I need to revert and release as a  major... so be it

PR #435, opened 2019-06-26
Remove excess assignment from docs

benjamingr:
Thanks for the project 🙏

This change removes an assignment of a variable to itself since Object.assign is already mutating.


benjamingr:
Thanks - I can't merge though :D

PR #444, opened 2019-08-07
Remove babel-polyfill

jelaxshan:
I have removed babel-polyfill in this PR, as it should be the responsibility of the package consumer to include polyfills. This avoids potential conflicts with other polyfills and removing this polyfill should also reduce bundle size. Additionally babel-polyfill is essentially just core-js under the hood, and there is no need to include both in this lib.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1054,6 +1054,7 @@
       "version": "6.26.0",
       "resolved": "https://registry.npmjs.org/babel-polyfill/-/babel-polyfill-6.26.0.tgz",
       "integrity": "sha1-N5k3q8Z9eJWXCtxiHyhM2WbPIVM=",
+      "dev": true,
       "requires": {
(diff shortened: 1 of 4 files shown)


redaxmedia:
I think this is a valid point but merging this should result in an major version bump.

wheresrhys:
I'll be discussing these compatibility problems with my co-maintainer @grug soon and we'll hopefully release a new version which keeps everybody happy.

wheresrhys:
[v8.0.0-alpha.1](https://www.npmjs.com/package/fetch-mock/v/8.0.0-alpha.1) attempts to address this issue. Please leave a :+1: on the pull request https://github.com/wheresrhys/fetch-mock/pull/457 if your problem is solved, or alternatively leave a comment @redaxmedia @jelaxshan

Issue #442, opened 2019-07-26
Feature request: Add option to match request body


gr2m:
Today, I can quite easily match the request body & request headers. Example

```js
fetchMock.postOnce(
  // match URL
  "https://api.github.com/app/installations/123/access_tokens",
  // response body
  {
    token: "secret123",
    expires_at: "1970-01-01T01:00:00.000Z",
    repositories: [{ id: 1 }, { id: 2 }, { id: 3 }]
  },
  {
    // match request headers
    headers: {
      accept: "application/vnd.github.v3+json",
      authorization: `bearer ${BEARER}`,
      "user-agent": "test",
      "content-type": "application/json; charset=utf-8"
    }
  }
);
```

It would be nice to be able to match the request body similar to request headers

```js
fetchMock.postOnce(
  // match URL
  "https://api.github.com/app/installations/123/access_tokens",
  // response body
  {
    token: "secret123",
    expires_at: "1970-01-01T01:00:00.000Z",
    repositories: [{ id: 1 }, { id: 2 }, { id: 3 }]
  },
  {
    // match request headers
    headers: {
      accept: "application/vnd.github.v3+json",
      authorization: `bearer ${BEARER}`,
      "user-agent": "test",
      "content-type": "application/json; charset=utf-8"
    },
    // match request body
    body: {
      permissions: {
        contents: "read"
      }
    }
  }
);
```

The current workaround is to use a matcher function

```js
fetchMock.postOnce(
  // match URL
  "https://api.github.com/app/installations/123/access_tokens",
  // response body
  {
    token: "secret123",
    expires_at: "1970-01-01T01:00:00.000Z",
    repositories: [{ id: 1 }, { id: 2 }, { id: 3 }]
  },
  {
    // match request headers
    headers: {
      accept: "application/vnd.github.v3+json",
      authorization: `bearer ${BEARER}`,
      "user-agent": "test",
      "content-type": "application/json; charset=utf-8"
    },
    // match request body using matcher function
    matcher(_, request) {
      expect(JSON.parse(String(request.body))).toStrictEqual({
        permissions: {
          contents: "read"
        }
      });
      return true;
    }
  }
);
```

With the matcher function  I need to provide my own assertion and it’s a little more code.

Apologies if this was requested before. I searched the docs and issues and couldn’t find anything about request body matching

wheresrhys:
@leggsimon was requesting this feature a few days ago too, and IIRC may have volunteered to implement it. Am I right?

[...]

PR #448, opened 2019-09-10
Match json bodies

leggsimon:
This PR allows for matching against JSON bodies. It adds the ability to add a `body` value in the options parameter which, when set, will match against requests with a `"Content-Type": "application/json"` header set.

### Example Usage

```js
fetchMock.mock('http://example.com/', 200, { body: { foo: 'bar' } })

fetch('http://example.com/', {
	method: 'POST',
	headers: {
		'Content-Type': 'application/json'
	},
	body: JSON.stringify({foo: 'bar'})
})

```

Addresses #442

### DIFF
--- a/docs/_api-mocking/mock_options.md
+++ b/docs/_api-mocking/mock_options.md
@@ -1,5 +1,5 @@
 ---
-title: "options"
+title: 'options'
 position: 1.3
 description: |-
   An object containing further options for configuring mocking behaviour
@@ -28,15 +28,24 @@ parameters:
       - Headers
     content: |-
       Match only requests that have these headers set
-    examples: 
+    examples:
       - |-
         {"Accepts": "text/html"}
+  - name: body
+    types:
+      - Object
+      - Headers
+    content: |-
+      Match only requests that have this body (requires a `Content-Type` request header to be set to `application/json`)
+    examples:
+      - |-
+        {"JSON": "body"}
   - name: query
     types:
       - Object
     content: |-
       Match only requests that have these query parameters set (in any order)
-    examples: 
+    examples:
       - |-
         {"q": "cute+kittenz", "format": "gif"}
   - name: params
@@ -44,7 +53,7 @@ parameters:
       - Object
     content: |-
       When the `express:` keyword is used in a string matcher, match only requests with these express parameters
-    examples: 
+    examples:
       - |-
         {"section": "feed", "user": "geoff"}
   - name: functionMatcher
--- a/docs/_api-mocking/mock_options.md
+++ b/docs/_api-mocking/mock_options.md
@@ -34,12 +34,11 @@ parameters:
   - name: body
     types:
       - Object
-      - Headers
     content: |-
-      Match only requests that have this body (requires a `Content-Type` request header to be set to `application/json`)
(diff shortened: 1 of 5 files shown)


docs/_api-mocking/mock_options.md
```diff
@@ -28,23 +28,32 @@ parameters:
...
     content: |-
       Match only requests that have these headers set
-    examples: 
+    examples:
       - |-
         {"Accepts": "text/html"}
+  - name: body
+    types:
+      - Object
+      - Headers
```
wheresrhys:
```suggestion
```

docs/_api-mocking/mock_options.md
```diff
@@ -28,23 +28,32 @@ parameters:
...
+    content: |-
+      Match only requests that have this body (requires a `Content-Type` request header to be set to `application/json`)
```
wheresrhys:
```suggestion
      Match only requests that send a JSON body with the exact same structure and properties as the one proviede here (also requires that the request has a `Content-Type` request header set to `application/json`)
```

docs/_api-mocking/mock_options.md
```diff
@@ -28,23 +28,32 @@ parameters:
...
+    examples:
+      - |-
+        {"JSON": "body"}
```
wheresrhys: Not sure what properties are better to use as an example, but in my experience using weighted terms like 'JSON' and 'body' leads to confusion i.e. 'So do I need to include a 'JSON' property that wraps my body?'

src/lib/generate-matcher.js
```diff
@@ -72,6 +73,23 @@ const getParamsMatcher = ({ params: expectedParams, matcher }) => {
...
+		if (method.toLowerCase() === 'get') {
+			// GET requests don’t send a body so the body matcher should be ignored for them
+			return true;
+		}
+
+		const lowerCaseHeaders = headerUtils.toLowerCase(
+			headerUtils.normalize(headers)
+		);
+
+		return lowerCaseHeaders['content-type'] === 'application/json' && body
```
wheresrhys: I wonder if this is a good idea to check for this. Even though it's bad practice to, there are circumstances where applications are sending json bodies without appropriate headers, and users could quite legitimately want to test those scenarios. Would there be any downsides to loosening this up to not require the header?

test/specs/routing.test.js
```diff
@@ -551,6 +551,90 @@ module.exports = fetchMock => {
...
+
+					await fm.fetchHandler('http://it.at.there/', {
+						method: 'POST',
+						body: JSON.stringify({ foo: 'woah!!!' }),
+						headers: { 'Content-Type': 'application/json' }
+					});
+					expect(fm.calls(true).length).to.equal(0);
+				});
+
+				it('should ignore the order of the keys in the body', async () => {
```
wheresrhys: 👌

docs/_api-mocking/mock_options.md
leggsimon: Ah yes, that’s a good point it’s not very clear. Will change

src/lib/generate-matcher.js
leggsimon: Having thought about it, I think you’re right, I think I’ll remove that requirement

wheresrhys:
Proper job

I will release it now

PR #451, opened 2019-09-27
Adds node-fetch to the peer dependencies

arcanis:
The `node-fetch` package is required here:

https://github.com/wheresrhys/fetch-mock/blob/master/src/server.js#L1

But isn't properly listed in the peer dependencies (it's only a dev dependencies). This may lead to non-deterministic installs, and will be purely disallowed in future Yarn releases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,11 @@
   "peerDependencies": {
     "node-fetch": "*"
   },
+  "peerDependenciesMeta": {
+    "node-fetch": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": ">=4.0.0"
   },
--- a/package.json
+++ b/package.json
@@ -53,6 +53,9 @@
     "path-to-regexp": "^2.2.1",
     "whatwg-url": "^6.5.0"
   },
+  "peerDependencies": {
+    "node-fetch": "*"
+  },
   "engines": {
     "node": ">=4.0.0"
   },


wheresrhys:
It's deliberately not included. See http://www.wheresrhys.co.uk/fetch-mock/#usagerequirements

wheresrhys:
Although I'm interested to know more about disallowing in future versions of Yarn. Are you able to share a reference for that?

arcanis:
> It's deliberately not included. See http://www.wheresrhys.co.uk/fetch-mock/#usagerequirements

If the peer dependency isn't required, then please list it as an [optional peer dependency](https://next.yarnpkg.com/configuration/manifest#peerDependenciesMeta.optional). It's supported by all three package managers (npm has a bug on `add`, but subsequent installs will properly omit the peer dependency requirement).

> Although I'm interested to know more about disallowing in future versions of Yarn. Are you able to share a reference for that?

This is part of what has been a long-term effort towards enforcing correctness in how packages list their dependencies. It started a year ago as an experimental mode called [Plug'n'Play](https://next.yarnpkg.com/features/pnp), and the results proved promising enough that the next major release (scheduled for the end of year) will enable it by default.

Note that omitting peer dependencies doesn't have purely theoretical problems - it also causes broken installs because the package managers need peer dependency information in order to know how to hoist the packages. When they're missing and you're not operating under PnP, you run the risk of either not being able to access them, or access the wrong one.

wheresrhys:
I didn't know about optional peer deps. Please modify this PR with the metadata expressing node-fetch's optional nature, then I think it makes sense to merge.

wheresrhys:
Actually, reading a bit more into it, I cannot see any mention of peerDependenciesMeta in the npm docs. It looks like a dependency must be optional or peer. Where have you seen that npm supports optional peer dependencies?

wheresrhys:
Have now found a [release log that mentions peerDependenciesMeta](https://npm.community/t/release-6-11-0/9572) - it's just the npm docs have not been updated

It's only supported in very recent npm, so now I worry that adding a peer dependency and using peerDependencyMeta will cause npm install to fail for projects using older npm and that do not use node-fetch.

Is optionalDependencies the best way out for now?

arcanis:
> Where have you seen that npm supports optional peer dependencies?

I've [implemented it](https://github.com/npm/cli/pull/224) 🙂

> I worry that adding a peer dependency and using peerDependencyMeta will cause npm install to fail for projects using older npm and that do not use node-fetch

It will only cause, at worst, a warning. Installs will exit with a success exit code.

> Is optionalDependencies the best way out for now?

Optional dependencies cannot replace optional *peer* dependencies; they are quite misnamed in that they are regular dependencies except that their build is allowed to fail.

wheresrhys:
> I've [implemented it](https://github.com/npm/cli/pull/224) 🙂
I guess you probably know what you're talking about in that case :joy:

OK, so if you add the peerDependenciesMeta with optional set to true in this PR, then I'm happy to merge

wheresrhys:
released as 7.5.0

wheresrhys:
@arcanis I wonder if you can help explain something to me

peerDependenciesMeta is not documented, but I note there is now an `optionalDependencies` property that is.

As you were the original implementer, do you know whether optionalDependencies is now the recommended approach?

arcanis:
`peerDependenciesMeta` is documented: https://yarnpkg.com/configuration/manifest#peerDependenciesMeta

`optionalDependencies` is an old field that doesn't do what you expect it to do; it just silences build errors (cf [doc](https://yarnpkg.com/configuration/manifest#optionalDependencies)).

PR #453, opened 2019-10-03
Check for Error.captureStackTrace before using it

### LINKED ISSUE #447: Regression: `Error.captureStackTrace` is not defined in non-Chromium browsers
Commit 81e011d7989f646bdcfb473322477a52004eba7b introduces the following code:

```js
class AbortError extends Error {
	constructor() {
		super(...arguments);
		this.name = 'AbortError';

		// Do not include this class in the stacktrace
		Error.captureStackTrace(this, this.constructor);
	}
}
```

However, `captureStackTrace` is a V8-ism and as a result this code will fail in non-Chromium based browsers. As [described on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types) this should be wrapped in a check for `captureStackTrace`:

```js
// Maintains proper stack trace for where our error was thrown (only available on V8)
if (Error.captureStackTrace) {
    Error.captureStackTrace(this, CustomError);
}
```

On my project tests are broken on CI when running on Firefox using fetch-mock 7.3.9, but work with 7.3.8.

birtles:
Error.captureStackTrace is a V8-ism so we should check if it exists
before trying to use it or else it will fail in non-Chromium browsers.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types

Fixes #447.

### DIFF
--- a/src/lib/fetch-handler.js
+++ b/src/lib/fetch-handler.js
@@ -11,7 +11,9 @@ class AbortError extends Error {
 		this.message = 'The operation was aborted.';
 
 		// Do not include this class in the stacktrace
-		Error.captureStackTrace(this, this.constructor);
+		if (Error.captureStackTrace) {
+			Error.captureStackTrace(this, this.constructor);
+		}
 	}
 }
 


wheresrhys:
Thanks for the PR. Will release as 7.5.1

PR #479, opened 2019-12-04
Signal request

wheresrhys:
Continued from https://github.com/wheresrhys/fetch-mock/pull/473

As implemented, it will be a breaking change for lots of people's tests because any expectation they've made on `options` will break if they are also using `Request`.

It looks like `new Request()` always creates a signal property (in the browser - not an issue in node-fetch). I can't expect users to always have to care about abort signals as an artefact even when they are not deliberately working with aborted fetches.

I somehow need to carry the signal in to the library on the options so it can be used successfully to abort, but avoid have default signals getting exposed on the options object when inspecting. Probably the answer is to pass signal around as a separate entity, but this involves a bit of work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,15 +40,15 @@
       "dev": true
     },
     "@types/estree": {
-      "version": "0.0.39",
-      "resolved": "https://registry.npmjs.org/@types/estree/-/estree-0.0.39.tgz",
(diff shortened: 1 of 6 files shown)


wheresrhys:
Turned out to be simpler than I thought as signal is not carried very deep into the library

PR #485, opened 2019-12-17
fix(abort): throwing DOMEXception when available

Sayan751:
As per spec when a fetch is aborted a DOMException is thrown with name 'AbortError'. However, the current implemnetation threw a custom error instance.  That should work for node.js due to unavailibility of DOMException. This commit ensures that when DOMException is available it is thrown instead.

Related: #416


wheresrhys:
Looks good - thanks

Issue #489, opened 2019-12-25
Implement optional partial body matching


wheresrhys:
Currently requests must match an entire JSON body

It seems valuable to allow an option `partialBodyMatching: Bool [default: false]` to match with similar semantics to [jest's objectContaining](https://jestjs.io/docs/en/expect#expectobjectcontainingobject) 

Could use https://www.npmjs.com/package/is-subset

wheresrhys:
@leggsimon you'd love to take this on wouldn't you?

leggsimon:
Yeah, could do. I’m not sure about the API though. Are you thinking that that `partialBodyMatching` option should be a property on the `matcher` object? For example:

```js
fetchMock.mock("http://example.com/", 200, {
  partialBodyMatching: true,
  body: {
    key1: "value1"
  }
});
```

Personally I think that interface seems a little clunky but I’m not sure if there’s a nicer option that wouldn’t be a breaking change.

Also, I would say it’s not immediately obvious, the above example would match the following bodies

```JSON
{ "key1": "value1" }
```

```JSON
{ "key1": "value1", "key2": "value2" }
```

But should it match: 

```JSON
{
	"foo": {
		"bar": {
			"key1": "value1",
			"key2": "value2"
		}
	}
}
```

wheresrhys:
yes to the API. But also it could be set as a global option so if a user is after partial matching then they can just set it once.

On the not immediately obviousness, it's how jest's object matching works to basically follow th object tree and for a match to occur the body would have to have the same value for the same property at the same depth. I can't remember what the array behaviour is

wheresrhys:
Decided I'd just do this. Quite easy in the end

PR #496, opened 2020-01-03
support body matching with Request objects

wheresrhys:
node-fetch support implemented so far

Harder in the browser as the body content cannot be accessed synchronously

I don't think I can fix this without a breaking change. this is because extracting the body content from a Request instance can only be done asynchronously using a spec compliant Request instance. Therefore the step to normalise a Request into the object fetch-mock passes around internally (the first thing that happens on handling a request) must be made async. This in turn means that the fetch handler cannot throw a synchronous error when something unexpected happens e.g. a call is made that has no corresponding mock.

It's unlikely that developers are relying on this behaviour other than when actively writing tests, but even so I think it'd be best to release as a major version. And even then I need to seriously weigh up what the best approach is:

- Always throw asynchronously, which is consistent, but may make debugging tests harder (or it might not... needs careful thought anyway)
- Refactor so that I don't need to throw asynchronously for most cases, but do for the body extraction case. This is probably a lot of work, will lead to a messier code base, and the inconsistency is likely to still confuse developers

... or something else??


test/specs/routing.test.js
```diff
@@ -580,6 +580,16 @@ module.exports = fetchMock => {
 					expect(fm.calls(true).length).to.equal(1);
 				});
 
+				it('should match when using Request', async () => {
+					fm.mock('http://it.at.there/', 200, { body: { foo: 'bar' } }).catch();
+
+					await fm.fetchHandler(new fm.config.Request('http://it.at.there/', {
```
renanvalentin: About your concerns of introducing async handling: It just feels natural using `async` / `await` since we are dealing with network requests that happens to be async by default.

test/specs/routing.test.js
```diff
@@ -763,8 +773,9 @@ module.exports = fetchMock => {
 		});
 
 		describe('unmatched calls', () => {
-			it('throws if any calls unmatched', async () => {
+			it.only('throws if any calls unmatched', async () => {
```
grug: We should probably get rid of the `only` and console.log from here

wheresrhys: Good spot. I'd like to add better linting to this project to catch stuff like this

test/specs/routing.test.js
wheresrhys: Yeah, I think the benefit of supporting ky outweighs the concern. It has to be a major release though so I will hold off while I consider whether there are other ways the API can be improved at the same time e.g. https://github.com/wheresrhys/fetch-mock/issues/489

renanvalentin:
can we merge this PR ?

test/specs/routing.test.js
wheresrhys: I'm looking at this again. Another thing that breaks when introducing this early asynchronicity is it is not possible to check synchronously that an expected call has been made

i.e. at the moment the following works:

```
myMethod() // which fetches 'http://alan.com'
expect(fetchMock.called('http://alan.com'))
```

but it will ned to be updated to 

```
await myMethod() // which fetches 'http://alan.com'
expect(fetchMock.called('http://alan.com'))
```
or

```
myMethod() // which fetches 'http://alan.com'
await fetchMock.flush()
expect(fetchMock.called('http://alan.com'))
```

While these may seem minor, it's potentially imposing lots of annoying little test rewrites on users. I'm not enjoying doing this for fetch-mock's own test suite, and don't particularly want to cause others that pain.

Still going to try to solve this in a nice way, but needs quite a bit of thought

wheresrhys:
I finally went for option2 - keep things synchronous except for this case. I found a neatish way to implement it - normalise request returns a promise, and then fetchHandler forks into as async/sync helper, as appropriate, depending on whether it detects a need to await a body

Issue #510, opened 2020-03-05
when using partial body matching, fetchMock.getOption is not there


keremciu:
I'm so excited about your changes in the library recently. partial body matching is a great one 🎊 

I updated library to 9.10 to use that feature but my tests are broken now.

```TypeError: Cannot read property 'getOption' of undefined``` error is this one.

How it's occurred? 

Basically I'm writing some tests like this one:

```
expect(
        fetchMock.called(/v1\/comments/, {
          method: "POST",
          body: {
            comment: {
              body: comment.body,
            }
          }
        })
      ).toBeTruthy();
```

If I remove body, it works because the problem is finding fetchMock instance: https://github.com/wheresrhys/fetch-mock/blob/master/src/lib/generate-matcher.js#L225

I've no idea how can I affect this generateMatcher code and put default fetchMock instance there as second argument:
https://github.com/wheresrhys/fetch-mock/blob/master/src/lib/inspecting.js#L10

What do you recommend for this case?

fetchMock.filterCalls is in same situation while it's related to same generateMatcher function

wheresrhys:
Fixed in v9.1.1

PR #511, opened 2020-03-05
passing fetchMock instance to generateMatcher calls

keremciu:
fix this issue: https://github.com/wheresrhys/fetch-mock/issues/510

generateMatcher needs fetchMock instance when there's a body key inside route object.


wheresrhys:
Thanks for the pull request. I've implemented a similar fix with tests also, so will merge that instead

Issue #528, opened 2020-04-08
feat: node-fetch's natural exports should be re-exported by mock


chet-manley:
The `node-fetch` package exports helpers like `Headers` and `Response` by default, allowing Node projects to use them in their code, usually like so:
```javascript
const { default: fetch, Headers } = require('node-fetch')
const headers = new Headers(...)
fetch('...', {headers})
```
If your code relies on these utilities, mocking `node-fetch` causes errors, such as: `TypeError: Headers is not a constructor`. It would be nice if `fetch-mock` (and `fetch-mock-jest` by association) re-exported these useful utilities, so as not to break production code. 

[Repl.it demonstrating issue](https://repl.it/repls/DarkvioletImmaterialIteration)

chet-manley:
My current work-around:
```javascript
// __mocks__/node-fetch.js
'use strict'

const nodeFetch = require('node-fetch')
const fetchMock = require('fetch-mock-jest')
const fetch = fetchMock.sandbox()

module.exports = exports = fetch
module.exports.default = exports
module.exports.Headers = nodeFetch.Headers
module.exports.Request = nodeFetch.Request
module.exports.Response = nodeFetch.Response

```

wheresrhys:
I think this is a good idea. Thanks for raising it.

There is however some nuance:fetch-mock is configurable to use _any_ implementation of fetch, and node-fetch is not the only one available (although it is the most popular), so I don't want to rush into this. Specifically, I wonder if I should export using proxies that can be used as getters for properties on fetchMock.config

Will think about it for a bit

chet-manley:
Reviewing your code, I see that the default behavior is to pull `Headers`, `Request`, and `Response` directly from `node-fetch` into your `config`, and allowing the user to overload each of them with their own implementation. A good compromise that leaves this behavior intact, would be to expose each of those classes either on the main `fetchMock` class or instance class provided by `FetchMock.createInstance`, pulling them from the config. Then, re-expose the sandbox class on a circular `default` property (`sandbox.default = sandbox`) to prevent breaking code that implements `node-fetch`'s export pattern. This adds useful functionality while preventing any breaking changes to your existing module.

wheresrhys:
I followed your reasoning there apart from the bit about sandbox. Could you put together a draft PR to demonstrate what you mean?

PR #532, opened 2020-04-12
feat: mock node-fetch's export pattern

chet-manley:
Resolves wheresrhys/fetch-mock#528

Modified `FetchMock`'s `sandbox` method to return an interface that more closely mocks `node-fetch`'s export pattern. This prevents code that relies on `node-fetch` from breaking during testing in some situations. Everything is done in the sandbox, so users that modify the main class config with their own implementations can still take advantage of the re-exposed `Headers`, `Request`, and `Response` classes if they so desire. 

Created test verifying the new return value.

### DIFF
--- a/src/lib/index.js
+++ b/src/lib/index.js
@@ -55,11 +55,17 @@ FetchMock.sandbox = function() {
 	const sandbox = Object.assign(
 		proxy, // Ensures that the entire returned object is a callable function
 		FetchMock, // prototype methods
-		this.createInstance() // instance data
+		this.createInstance(), // instance data
+		{
+			Headers: this.config.Headers,
+			Request: this.config.Request,
+			Response: this.config.Response
+		}
 	);
 
 	sandbox.bindMethods();
 	sandbox.isSandbox = true;
+	sandbox.default = sandbox;
 	return sandbox;
 };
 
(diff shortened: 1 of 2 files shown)


chet-manley:
As an aside, your dev environment fails to run in Node v13.12 (and likely >=13.2). Since your `package.json` is configured with `type: module`, Node throws when attempting to `require()` any file with a `.js` extension. This breaks basically everything in your `Makefile`. I changed the type to `commonjs` to run linting and tests, and noted the tests threw 2 `UnhandledPromiseRejectionWarning` and 1 `DeprecationWarning: Buffer()`. 

Also, not sure if it's on purpose or not, but `npm test` only runs linting.

wheresrhys:
I think the answer to the module vs commonjs issue is to put a package.json with `type: commonjs` in the src and test directories, but thanks for flagging it.

I never use npm scripts - it's possible some piece of tooling I've used in the past added that without me noticing. I'll point it to something sensible.

Thanks for the PR - it looks like a good solution

wheresrhys:
Very odd that the linting has suddenly stopped passing. It fails on node 12 too, and worked very recently

chet-manley:
In researching a similar but unrelated issue, I discovered Node v12.16 followed in the footsteps of v13.2 and changed all of the ESM loading warnings to errors. It's obviously a breaking change, so not sure why it was allowed into v12, but it's the most likely cause of the issue.

PR #540, opened 2020-05-07
Allow query matcher to use repeated query string parameters

lodrantl:
Pretty self explanatory PR.

```
fetchMock.mock({
  query: {
    a: ["b", "c"]
  }
}....
```
which will match:
http://my.domain.url/api?a=b&a=c

But not
http://my.domain.url/api?a=b
http://my.domain.url/api?a=b&a=c&a=d

### DIFF
--- a/docs/_api-mocking/mock_matcher.md
+++ b/docs/_api-mocking/mock_matcher.md
@@ -146,7 +146,7 @@ parameters:
         types:
           - Object
         content: |-
-          Match only requests that have these query parameters set (in any order)
+          Match only requests that have these query parameters set (in any order). In order to match keys which are used multiple times use an array of strings.
         examples:
           - |-
             {"q": "cute+kittenz", "format": "gif"}
--- a/docs/_api-mocking/mock_matcher.md
+++ b/docs/_api-mocking/mock_matcher.md
@@ -150,6 +150,8 @@ parameters:
         examples:
           - |-
             {"q": "cute+kittenz", "format": "gif"}
+          - |-
+            {"tags": ["cute", "kittenz"]}
       - name: params
         types:
           - Object
--- a/src/lib/matchers.js
+++ b/src/lib/matchers.js
@@ -87,13 +87,15 @@ const getQueryStringMatcher = ({ query: expectedQuery }) => {
 		debug('  Actual query parameters:', query);
 		return keys.every((key) => {
 			const value = query[key];
-			const expectedValue = query[key];
+			const expectedValue = expectedQuery[key];
 			if (Array.isArray(value) && Array.isArray(expectedValue)) {
-				return value.every(q => expectedValue.includes(q)) &&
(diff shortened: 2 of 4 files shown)


wheresrhys:
Interesting. Question - what should be the behaviour for `http://my.domain.url/api?a[]=b&a[]=c&a[]=d` ?

lodrantl:
Are you asking if  `http://my.domain.url/api?a[]=b&a[]=c&a[]=d`
would be accepted for matcher
```
fetchMock.mock({
  query: {
    a: ["b", "c", "d"]
  }
}....
```
Right now it is not, I think PHP would accept it right?
As far as a mocking framework goes you could always use
```
fetchMock.mock({
  query: {
    "a[]": ["b", "c", "d"]
  }
}....
```

wheresrhys:
Yeah quite right - just been reading up on the lack of any spec on multi-value query strings, so keeping the behaviour simple and just treating all keys as strings is right.

docs/_api-mocking/mock_matcher.md
```diff
@@ -150,6 +150,8 @@ parameters:
         examples:
           - |-
             {"q": "cute+kittenz", "format": "gif"}
+          - |-
+            {"tags": ["cute", "kittenz"]}
```
wheresrhys, line 154: Could you update the docs just above to say: 
> Match only requests that have these query parameters set (in any order). In order to match keys which are used multiple times use an array of strings.

src/lib/matchers.js
```diff
@@ -85,7 +85,17 @@ const getQueryStringMatcher = ({ query: expectedQuery }) => {
 		const query = querystring.parse(getQuery(url));
 		debug('  Expected query parameters:', expectedQuery);
 		debug('  Actual query parameters:', query);
-		return keys.every((key) => query[key] === expectedQuery[key]);
+		return keys.every((key) => {
+			const value = query[key];
+			const expectedValue = expectedQuery[key];
+			if (Array.isArray(value) && Array.isArray(expectedValue)) {
```
wheresrhys, line 91: What about if my mock is `key: ['val']`. I _think_ this won't match `?key=val` as the parsed query string will be `{key: 'val'}`

So needs a test for that scenario. And I think the clearest way to write the matcher is to coerce both value and expectedValue to arrays up front, then compare the arrays

lodrantl: True, I think it should be okay now. Looks a bit cleaner too.

ayZagen:
I believe #555 covers this also.

wheresrhys:
@ayZagen I think I will merge this PR as the tests cover an interesting edge case, and I think the object comparison is nicer - hopefully it is not too much work to rebase onto it

@lodrantl  - thanks for the changes. Will release as v9.7.0

ayZagen:
@wheresrhys This would fail with query params containing `null` and `boolean` values

Issue #554, opened 2020-05-12
URI encoded query parameters are not matched


rvinzent:
Hi, thanks for the package!

In the below code snippet I would expect at least one of the fetch mock endpoints to match successfully, but I am unable to make the match work. I have tried using the `query` match option as well as matching the full string URL, and the `URL` instance directly, but nothing seems to make my test pass. 

In the console output you can see that the logged actual URL and warnings for unmatched URL are identical. Is this a bug or am I missing something obvious?

[REPL](https://repl.it/repls/DimgreyUnrulyJavabytecode)

``` javascript
const { DateTime }  = require('luxon')
const fetch = require('isomorphic-fetch')
const fetchMock = require('fetch-mock')

const mockResponse = [{ id: 1, prop: 'value' }]
const datetime = DateTime.local()

const url = new URL('http://fake.com/api/any')
url.searchParams.append('date', datetime.toString())

fetchMock.getOnce(url, 200)
fetchMock.getOnce(url.toString(), 200)
fetchMock.getOnce('http://fake.com/api/any', {
  status: 200,
  query: { date: datetime.toString() }
})

fetch(url.toString()).then(response => {
  console.log(fetchMock.done())
})

setTimeout(() => { console.log('url: ', url.toString()) }, 1000)
```

Here is the console output from the REPL:
```
Warning: http://fake.com/api/any?date=2020-05-12T18%3A07%3A23.852-00%3A00 not called
Warning: http://fake.com/api/any?date=2020-05-12T18%3A07%3A23.852-00%3A00 not called
Warning: http://fake.com/api/any not called
false
url:  http://fake.com/api/any?date=2020-05-12T18%3A07%3A23.852-00%3A00
```

The logged URL and the URL displayed in the warning are identical.

----
EDIT: I am trying desperately to add a passing call in the REPL but it refuses to work. However, I see similar behavior in my project's tests when `fetch` is actually able to match successfully in other cases. Unfortunately I am unable to share the full code, and there is a bit more happenning under the hood, but the problem remains where we have both an "unmatched call" and a warning for an extra call on the same URL.

Here is the fetch mock setup, which I have also tried using the `query` match property: 
```javascript
    fetchMock.getOnce(`end:/api/any?dateParam=${expectedIsoString}`, {
      status: 200,
      body: JSON.stringify(mockResponse),
    })
```

Additionally, this test case passes fine, so I can be fairly sure this is a result of the query string encoding:

```javascript
[...]

PR #555, opened 2020-05-12
fix: request with query params not containing string values not validated correctly

ayZagen:
Given the example: 
```js
const mocker = fetchMock.sandbox()

const params = {
   a: 1,
   b: 2
}

const request = mocker.mock({
  url: 'http://test.example.com',
  method: 'GET',
  query: params
}, 200);

fetch( {
   url : 'http://test.example.com?'+ qs.stringify(params),
   method: 'GET'
}).then( () => {
   assert request.called() === true   // fails
})
```

The problem is query matcher checks for passed query values without transforming them. For ex. if the passed query param is a number it will fail because received query params will be string.

### DIFF
--- a/docs/_api-mocking/mock_matcher.md
+++ b/docs/_api-mocking/mock_matcher.md
@@ -166,10 +166,10 @@ parameters:
         types:
           - Object
         content: |-
-          Match only requests that have these query parameters set (in any order)
+          Match only requests that have these query parameters set (in any order). Query parameters are matched by using Node.js [querystring](https://nodejs.org/api/querystring.html) module. If it doesn't fulfil your requirements, you should append query parameters to the url manually.
         examples:
           - |-
-            {"q": "cute+kittenz", "format": "gif"}
+            {"q": "cute+kittenz", "format": "gif", inform: true}
       - name: params
         versionAdded: 6.0.0
         types:
--- a/src/lib/compile-route.js
+++ b/src/lib/compile-route.js
@@ -1,3 +1,5 @@
+const querystring = require('querystring');
+
 const { debug, setDebugNamespace, getDebug } = require('./debug');
 
 const isUrlMatcher = (matcher) =>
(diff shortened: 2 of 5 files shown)


ayZagen:
Also related to #554

wheresrhys:
I can see this being a convenience for developers, even though query strings, technically, can only ever be strings, so I'm in favour of the addition.

I think it is however a bit odd to coerce numbers but not boolean and null, and it also raises the question (which must be documented/implemented) of what to do with keys that exist but contain undefined.

In summary - if we start coercing _some_ values we should coerce other values where it's sensible to do so, and have clear documentation of when coercion is not applied.

I'd favour
- Booleans are coerced
- objects (including null and functions) throw an error
- undefined is treated the same as a non-existent key

Questions:
- Does that sound sensible?
- Can you think of any omissions?
- Are you happy to implement that behaviour in this PR?

ayZagen:
You are right about other types. But regarding to your proposals I believe leaving to a module rather than implementing our own ways for dealing with it would cover most cases. We could also add a global option for letting users to use their own module. One of the possible module which users could choose is `qs` and it has really edge case difference with `querystring` module. IMHO In this state, updating type definition of query to `object` would be enough.

ayZagen:
In addition, as you know `querystring` is a core module in Node.js and it is only serving this purpose and I believe it sticks to common URI semantics. Implementing other ways would confuse users. If users have other requirements they always can use full url matching ( which is I am currently doing right now)

ayZagen:
> We could also add a global option for letting users to use their own module

Or we could provide an interface for passing arguments to [querystring](https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options)

wheresrhys:
Ok, so it seems querystring covers most cases in a pretty intuitive way. If you wanna expand the tests & types to cover these extra types of input (boolean & undefined I think are particularly important... the others less so) I'll take a proper look at your PR when that's done.

ayZagen:
added test cases as you requested

wheresrhys:
Closing in favour of #557, which tweaks the docs and tests a bit

PR #557, opened 2020-05-13
Fix/req with query not matched

wheresrhys:


wheresrhys:
Thanks for your contribution @ayZagen . Will be released as v9.9.0

PR #564, opened 2020-05-21
make response available to be inspected

wheresrhys:
@grug I couldn't resist implementing. Still needs tests but expect it to be released soon.

The api is as simple as can be - exposing `response` on each item of the array returned by `calls()`


grug:
Brilliant stuff!

wheresrhys:
done - will be released shortly

grug:
This is super awesome stuff! Very keen to use it :)

PR #587, opened 2020-08-16
Detached route creation

wheresrhys:
TODO
- [ ] Tests that methods behave basically the same as existing methods
- [ ] Test that inspection based on passing in routes basically works
- [ ] Deprecate (with warning) use of route name
- [ ] Document in the midst of existing docs
- [ ] New docs page outlining the new approach

### DIFF
--- a/src/Route/index.js
+++ b/src/Route/index.js
@@ -10,8 +10,7 @@ const isUrlMatcher = (matcher) =>
 const isFunctionMatcher = (matcher) => typeof matcher === 'function';
 
 class Route {
-	constructor(args, fetchMock) {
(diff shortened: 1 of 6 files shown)


src/Route/index.js
```diff
@@ -10,8 +10,7 @@ const isUrlMatcher = (matcher) =>
 const isFunctionMatcher = (matcher) => typeof matcher === 'function';
 
 class Route {
-	constructor(args, fetchMock) {
-		this.fetchMock = fetchMock;
+	constructor(args) {
```
wheresrhys, line 13: go back to ...args now we can

PR #590, opened 2020-09-14
Update v6-v7-upgrade-guide.md

joekrump:
Fix a small typo in docs.


joekrump:
I really appreciate these docs as well as the ones for v5 to v6 upgrade. They helped when I needed to do the update due to v5 depending on an old, version of `node-fetch` to was affected by [this CVE](https://github.com/advisories/GHSA-w7rc-rwvf-8q5r).

wheresrhys:
thanks for the PR and the feedback - glad somebody uses the old guides

Issue #576, opened 2020-06-29
Missing type for `lastResponse`


ashoksudani:
Thanks for this beautiful library.

Just wanted to point that it does not have type for `lastResponse` in defined typescript types.

![Screenshot 2020-06-29 at 2 46 55 PM](https://user-images.githubusercontent.com/552975/85982086-d0e67600-ba17-11ea-9cfa-5576eacddaad.png)

wheresrhys:
Please submit a PR to modify the types.

PR #591, opened 2020-10-01
Add type definition for lastResponse

Zoe7:
This solves https://github.com/wheresrhys/fetch-mock/issues/576 🙂

### DIFF
--- a/types/fetch-mock-tests.ts
+++ b/types/fetch-mock-tests.ts
@@ -123,6 +123,15 @@ lastOptions = fetchMock.lastOptions("http://test.com", {
     method: "GET",
 });
 
+let lastResponse: Response = fetchMock.lastResponse();
+lastResponse = fetchMock.lastResponse(true);
+lastResponse = fetchMock.lastResponse("http://test.com");
+lastResponse = fetchMock.lastResponse(/https?:\/\/test.com/);
+lastResponse = fetchMock.lastResponse("http://test.com", "GET");
+lastResponse = fetchMock.lastResponse("http://test.com", {
+    method: "GET",
+});
+
 fetchMock.get("http://test.com", 200);
 fetchMock.getOnce("http://test.com", 200);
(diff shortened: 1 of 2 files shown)


wheresrhys:
Thanks for this. Could you add something in https://github.com/wheresrhys/fetch-mock/blob/master/types/fetch-mock-tests.ts that verifies that this type is defined as well please.

If you pull from master the linting errors in the build should be fixed, and I think the failed test job was just a blip in the coveralls service

Zoe7:
Done! (I rebased on master to get the latest changes)

I'm a bit puzzled by the CI checks. 
The linting issue was obvious enough from the CI error, but somehow I never managed to reproduce it locally (I'm assuming my local environment just isn't set up properly, and I'm running on Windows so running a Makefile there can be a little annoying).
The test one I don't get either. I highly doubt my changes would have anything to do with that, and yet it failed again 🤷‍♀️

I can also see some types checks are failing on master now, but I can't reproduce that locally either (I was thinking I could help you fix that if you're not super familiar with typescript, but it's hard when stuff seem to work locally).

Zoe7:
Got my hands on a mac and tried running the linting there (prior to your fix) and it didn't fail with the same errors as on CircleCI either.
I cloned the project, `npm install` and then `make lint-ci` and I'm running node 12 which is also the one that CI should be using for this check. So I'm really quite puzzled by this.

Is that something you're aware of and have any clue about? I'd be happy to investigate further otherwise, now that my curiosity has been piqued 😄

wheresrhys:
Code looks good - it may be that code coverage only works if the build has access to credential sfor coveralls, and if it's kicked off via a forked repo maybe it doesn't get that.

Will merge and debug the build failure if I need to.

Thanks again for the fix

Zoe7:
>Thanks again for the fix

You're welcome! 🙂

PR #592, opened 2020-10-06
Fix: support overwriting routes with regex matcher

adamgruber:
I came across this issue while trying to mock an existing route using regex matchers.

Prior to this change, using a regex matcher combined with `overwriteRoutes: true` would not overwrite the route. This is because the inspection was comparing regexes with `===` which will always return `false`. To fix this, object identifiers are coerced to strings before comparing. Function matchers are not affected by this change.

As a side note, if you agree with this change and accept this PR, would you mind adding the `hacktoberfest-accepted` label?

### DIFF
--- a/src/lib/set-up-and-tear-down.js
+++ b/src/lib/set-up-and-tear-down.js
@@ -13,11 +13,13 @@ FetchMock.mock = function (...args) {
 FetchMock.addRoute = function (uncompiledRoute) {
 	debug('Adding route', uncompiledRoute);
 	const route = this.compileRoute(uncompiledRoute);
-	const clashes = this.routes.filter(
-		({ identifier, method }) =>
-			identifier === route.identifier &&
-			(!method || !route.method || method === route.method)
-	);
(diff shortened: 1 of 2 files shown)


adamgruber:
The tests are passing but the `ci/circleci: test` check appears to have failed with this error:
```
/home/circleci/project/node_modules/coveralls/bin/coveralls.js:19
      throw err;
      ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
(Use `node --trace-uncaught ...` to show where the exception was thrown)
Makefile:39: recipe for target 'coverage' failed
make: *** [coverage] Error 1
```

wheresrhys:
Apologies for the delay - I was trying to figure out whether or not I thought this would be a breaking change. I think it's safe to release though.

Thanks for your contribution

wheresrhys:
On adding the hacktoberfest label - sorry but I don't think hacktoberfest should be using a mechanism which clutters up open sourcers' metadata. Hopefully you can submit this some other way.

PR #593, opened 2020-10-23
Fix/protocol relative urls

carlosescri:
Hi @wheresrhys, thank you for your amazing work with this library.

I found a problem while mocking calls from a client library that works in the browser and sends protocol-relative requests.

I was trying the simplest mock possible with the `any()` shortcut, and it was throwing an error like this:

image

Seems that the `URL()` constructor does not accept protocol-relative URLs without a base, like `'http://dummy'` (at least in Chrome) so I split the regex in two to handle that case by setting the same dummy base as for the last case in the code, but returning the href as in the original case.

The relevant code that breaks is here:

https://github.com/wheresrhys/fetch-mock/blob/c553baf8295d2a57ad05b4d7524b866a42a0a9e6/src/lib/request-utils.js#L19-L34

I hope I did it right 😱

### DIFF
--- a/src/lib/request-utils.js
+++ b/src/lib/request-utils.js
@@ -1,7 +1,8 @@
 let URL;
 // https://stackoverflow.com/a/19709846/308237
-// modified, URL constructor does not support protocol-relative urls
+// split, URL constructor does not support protocol-relative urls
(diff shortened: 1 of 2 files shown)


wheresrhys:
Yep perfect - thanks for the fix

PR #594, opened 2020-11-17
Fix types for mock's aliases

lourenci:
All the mocking API functions should also accept an object as an argument in the `matcher` parameter like in the `mock` function.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -223,7 +223,7 @@ declare namespace fetchMock {
          * @default true
          */
         includeContentLength?: boolean;
-        
+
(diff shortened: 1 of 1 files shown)


wheresrhys:
Hi @lourenci , and thanks for the contribution

This will fail when it hits master because there is a problem with dtslint validation but CI does not run on pull requests from other users at the moment (because of a setting somewhere I need to find and change).

You seem to know typescript. If you could run `dtslint --expectOnly types` locally and see if it passes that would be great. If you can figure out how to fix the type validation errors even better

PR #607, opened 2021-05-08
Accept data: URLs

kdau:
Browsers have supported data: URLs in the fetch API for some time now. node-fetch will support them as well in its upcoming v3. The fetch API is the simplest and most reliable way to extract files from data: URLs.

Adding this support precludes a future "data:" matching prefix. This change should not break any existing HTTP(S) matchers, but since the "data:" scheme was previously stripped from URLs passed to matchers, this change will break any existing matchers for data: URLs that were full-string, "begin:", front-anchored regex/glob, or otherwise assumed that the "data:" scheme would be absent.

I did not include support for blob: URLs since they are opaque, less feasible to meaningfully test, and not supported by node-fetch.

### DIFF
--- a/src/lib/request-utils.js
+++ b/src/lib/request-utils.js
@@ -1,7 +1,7 @@
 let URL;
-// https://stackoverflow.com/a/19709846/308237
+// https://stackoverflow.com/a/19709846/308237 plus data: scheme
 // split, URL constructor does not support protocol-relative urls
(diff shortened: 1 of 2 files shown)


wheresrhys:
Thanks for working on this, and sorry for the very slow response. I'll pull it into fetch-mock@10

PR #610, opened 2021-05-15
FIX: fetch-mock Response.body is not always a stream

### LINKED ISSUE #609: fetch-mock Response.body is not always a stream in Node.js
[WhatWG fetch API spec](https://fetch.spec.whatwg.org/#body-mixin) requires that Body.body is of type ReadableStream. node-fetch [interprets the standard](https://github.com/node-fetch/node-fetch#bodybody) as requiring a Node.js Readable stream. fetch-mock [returns a Buffer](https://github.com/wheresrhys/fetch-mock/blob/4b54e6db42a3e87700ef37a27a7a4d7df29cb3d7/test/server-specs/server-only.test.js#L16) if anything other than a Readable is given to the Response object.

In Node:

```javascript
const fetchMock = require('fetch-mock');
const fetch = fetchMock.sandbox();

fetch.get('http://www.www.com.com', 'woo woo woo');

fetch('http://www.www.com.com').then(resp => resp.body.pipe(process.stdout));
```

...causes:

```
(node:36650) UnhandledPromiseRejectionWarning: TypeError: resp.body.pipe is not a function
    at /Users/fluggo/software/fetchmocktest/test-stream.js:6:56
```

This appears to be a bug in the ResponseBuilder, I have a patch.

fluggo:
The WhatWG fetch spec requires that Body.body (and therefore Response.body) is always a readable stream, but it doesn't always happen that way in Node. This seems to be because ResponseBuilder tries to fetch the stream module from fetchMock, but it looks in the wrong place.

Fixes #609.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,5 @@
 exclude_patterns:
 - docs
 - test/specs
+- test/client-specs
+- test/server-specs
--- a/src/lib/response-builder.js
+++ b/src/lib/response-builder.js
@@ -137,9 +137,12 @@ e.g. {"body": {"status: "registered"}}`);
 
 		// On the server we need to manually construct the readable stream for the
 		// Response object (on the client this done automatically)
-		if (this.Stream) {
+		if (
+			this.fetchMock.Stream &&
+			!(this.body instanceof this.fetchMock.Stream.Readable)
+		) {
 			this.debug('Creating response stream');
-			const stream = new this.Stream.Readable();
+			const stream = new this.fetchMock.Stream.Readable();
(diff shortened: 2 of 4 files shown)


fluggo:
The CI check that shows failure if code is similar _in test cases_ is not fun.

fluggo:
I've fixed all the issues I can find. I can't access the circleci tests to see what it's complaining about.

wheresrhys:
In the latest fetch-mock@10 I just use native fetch and I think all these issues should now be resolved.

Thanks for working on this, and apologies for the very slow response.

Issue #608, opened 2021-05-15
fetch-mock does not forward requests correctly when jsdom-global is in effect


fluggo:
fetch-mock has a function [patchNativeFetchForSafari](https://github.com/wheresrhys/fetch-mock/blob/2f6464c5a678ed7dd25e54d4956d0ea7088c1a3e/src/lib/fetch-handler.js#L24) that wraps the native fetch when it detects it's running in Safari. By a funny coincidence, what fetch-mock looks for is exactly how [jsdom identifies itself](https://github.com/jsdom/jsdom/blob/9f6644680230c59db02935ccb131ab73f1d802ac/lib/jsdom/living/navigator/NavigatorID-impl.js#L30).

The side effect is that, when jsdom-global is loaded, the patchNativeFetchForSafari function only passes the first argument on to the nativeFetch function. Like so:

```javascript
require('jsdom-global/register');
const fetchMock = require('fetch-mock');

fetchMock.config.fetch = function() {
  console.log(arguments);
  return Promise.reject(new Error('nyeh'));
}

const sandbox = fetchMock.sandbox();
sandbox.spy(true);

sandbox('http://localhost:2021', { headers: { 'cookie': 'id=5' } }).catch(() => {});
```

Produces:

```console
fluggo@NONEYABIZ:~/software/fetchmocktest$ node test.js 
[Arguments] { '0': 'http://localhost:2021/' }
```

...leaving out the specified header.

Either the detection needs to be more specific, nativeFetch should be called as `nativeFetch.apply(arguments)`, or both. For my purposes, my workaround will probably be `navigator = { vendor: 'Compy' };`.

wheresrhys:
The safari hack is now removed

PR #611, opened 2021-05-15
FIX: Ensure all arguments are passed to base fetch function

fluggo:
This is not a great fix; it ignores that method is only inspected when it's in the first fetch() parameter, and it glosses over the fact that the navigator object, for historical reasons, is [very constrained in the responses it gives](https://html.spec.whatwg.org/multipage/system-state.html#client-identification) (which is why the code snagged jsdom in the first place).

It does, however, solve my immediate need of getting fetch-mock to interoperate with jsdom.

Fixes wheresrhys/fetch-mock#608


wheresrhys:
The safari patch is removed from latest fetch-mock so no longer needed. Thanks for working on this and sorry for the slow response

Issue #616, opened 2021-07-27
Provide RequestInit object to native fetch on Safari


lar-dragon:
https://github.com/wheresrhys/fetch-mock/blob/4b54e6db42a3e87700ef37a27a7a4d7df29cb3d7/src/lib/fetch-handler.js#L34

This function expects Request object, but it may be a two arguments call witch string and RequestInit object.

lar-dragon:
If you want avoid this issue, normalise request object before call fetch by simple function:

```typescript
(...query: Parameters<typeof window.fetch>) => new Request(...query)
```

PR #617, opened 2021-07-27
Allow RequestInit in native fetch on Safari

lar-dragon:
Initialise Request object if first argument is string.

#616


wheresrhys:
Thanks for the fix. In the latest fetch-mock I've removed this entire hack targetting Safari though, so closing this as I won't be doing any work on earlier versions of fetch-mock

PR #622, opened 2021-11-08
Support exact-matching on FormData bodies

kylekthompson:
Hello 👋 

While attempting to mock some client-side calls to `fetch` using `fetch-mock` 9.11.0, I was hoping I'd be able to match `FormData` bodies in addition to JSON. I do see in the [docs]() that it's called out that JSON is supported. Would you be open to a change supporting other kinds of bodies?

The diff as-is will support exact matches of `FormData` (see the included tests; I've also patched this change locally in my repository as a test and validated that it solves my issues there as well). It _also_ will likely support other exact-matches of bodies (e.g. plaintext), but I've not explicitly called out support for that. I'd be happy to do some work to limit it to `FormData` only if you'd like (I'll just need to investigate a little more to make it compatible with both client and server environments).

### DIFF
--- a/test/specs/routing/body-matching.test.js
+++ b/test/specs/routing/body-matching.test.js
@@ -1,3 +1,4 @@
+const FormData = require('form-data');
 const chai = require('chai');
 const expect = chai.expect;
 
@@ -64,7 +65,7 @@ describe('body matching', () => {
 		expect(fm.calls(true).length).to.equal(0);
 	});
 
-	it('should not match if body sent isn’t JSON', async () => {
+	it('should not match if the bodies are of different types', async () => {
 		fm.mock({ body: { foo: 'bar' } }, 200).catch();
 
 		await fm.fetchHandler('http://a.com/', {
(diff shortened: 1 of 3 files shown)


test/specs/routing/body-matching.test.js
```diff
@@ -112,6 +113,54 @@ describe('body matching', () => {
...
+		actualFormData.append('fizz', 'buzz');
+		await fm.fetchHandler('http://a.com/', {
+			method: 'POST',
+			body: actualFormData,
+			headers: {},
+		});
+		expect(fm.calls(true).length).to.equal(0);
+	});
+
+	it('should not ignore the order of the form fields', async () => {
```
kylekthompson, line 146: We might be able to do something more complicated here, but I was a touch worried about the semantics of having multiple values for the same keys/arrays/etc.

We'd probably want to convert the `FormData` objects to JS objects and compare those as we are today (with `isEqual`) instead?

PR #647, opened 2022-10-29
types: added global URL type to `MockMatcherUrl`

flevi29:
In the [docs](https://www.wheresrhys.co.uk/fetch-mock/#api-mockingmock_matcher) the second entry to `Argument values` states that `url` can be any of `String` and `URL`, however the types do not support `URL`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,12 +72,12 @@
     "eslint-config-origami-component": "1.0.0",
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-prettier": "^2.6.1",
-    "karma": "^3.1.4",
+    "karma": "^6.4.1",
     "karma-chai": "^0.1.0",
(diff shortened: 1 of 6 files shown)


flevi29:
It seems that `karma-webpack 3.0.5` needs `webpack@"^2.0.0 || ^3.0.0"`, but the current version sits at `4.46.0`. NPM is strict with peers. There are 3 options:
1. Use `--legacy-peer-deps` when installing dependencies (dirty, peer deps are enforced for a good reason in my opinion)
2. Update karma
3. Downgrade webpack

flevi29:
I am not sure what to do about the last issue, tests seem to pass alright. It's a `coveralls` issue, token related or something similar. Can you do something about it @wheresrhys ?

wheresrhys:
Thanks for this. I will pull this change into fetch-mock@10

PR #680, opened 2024-06-17
fix: update types to be compatible with v10

wolfy1339:


wheresrhys:
Thanks for this

wheresrhys:
Sorry, but there was something about your commit that also broke the types
for the use cases the CI covers.

Can you please share more details of your setup, ideally a small repo which
installs fetch-mock and has the typescript config you use

wolfy1339:
The CI case handles CJS, and can be updated for dual-support. I get that it broke, it can be fixed

We use `fetch-mock` all over the @octokit org for the JS packages

Specifically this repo: https://github.com/octokit/core.js/pull/686
It might not be the smallest, but it show the issue and how the issue is not present before

The error I am getting is the following:
```
test/issues.test.ts:2:8 - error TS1192: Module '"/home/runner/work/core.js/core.js/node_modules/fetch-mock/types/index"' has no default export.
```
Since the types only have CJS exports, TS complains. Runtime is fine

wolfy1339:
I created a minimal reproducible test case, https://stackblitz.com/edit/fetch-mock-680?file=index.ts

wheresrhys:
Thank you

wheresrhys:
I think this has fixed the tests and I've reverted to your original change too https://github.com/wheresrhys/fetch-mock/blob/main/types/fetch-mock-tests.ts#L1C1-L1C28

Thanks for the pointers

PR #732, opened 2024-07-22
docs: add Int as response parameter type of mock function

syumai:
[`Int` is an allowed response type for status codes](https://www.wheresrhys.co.uk/fetch-mock/docs/fetch-mock/API/Mocking/Parameters/response/#status-code), but was missing from the mock function documentation.


wheresrhys (approved):
Thanks for this

PR #764, opened 2024-08-01
feat(matchers): add missingHeaders

### LINKED ISSUE #716: `headers` are not matched strictly
Hello I noticed that when using `toHaveFetched(url, { headers: {} })`, it will always pass no matter what headers are passed but it should pass only if no headers are passed

```js
    fetchMock.get('/api', { body: {} })
// custom wrapper around fetch
    await api.get('2')
    expect(fetchMock).toHaveFetched('/api/2', {
      headers: {
        // no Accept
        Accept: 'application/json'
        // 'Content-Type': 'application/json',
      },
    })
```

posva:
Close #716

### DIFF
--- a/packages/core/src/__tests__/Matchers/headers.test.js
+++ b/packages/core/src/__tests__/Matchers/headers.test.js
@@ -75,6 +75,20 @@ describe('header matching', () => {
 		).toBe(false);
 	});
 
+	it('not error when request sent without headers', () => {
+		const route = new Route({
+			missingHeaders: ['a'],
+			response: 200,
+		});
+
+		expect(
+			route.matcher({
+				url: 'http://a.com/',
+				options: {},
+			}),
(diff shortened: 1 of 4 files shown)


packages/core/src/__tests__/Matchers/headers.test.js
```diff
@@ -45,6 +45,36 @@ describe('header matching', () => {
...
+		});
+		expect(
+			route.matcher({
+				url: 'http://a.com/',
+				options: {
+					headers: { a: 'b' },
+				},
+			}),
+		).toBe(false);
+	});
```
wheresrhys, line 76: could do with an extra test "not error when request sent without headers"

wheresrhys (changes requested):
Implementation looks great.

Just left one request for an additional test, and it'd be good if you could also update the docs

Issue #760, opened 2024-07-30
Export more types from @fetch-mock/core


brucespang:
Hi!

First of all, really excited that you've made some changes to fetch-mock lately! I'm looking forward to upgrading us to 10.x, and have run into some issues with types. I'm very happy to submit a PR, but would like to know whether the change is intentional or not before I do.

Previously, we were using fetch-mock with:

```typescript
import { MockResponse, MockResponseFunction, MockRequest, MockResponseObject } from 'fetch-mock';
```

With 10.1.1, this gives the error `Module '".."' has no exported member 'MockResponse'. Did you mean to use 'import MockResponse from ".."' instead?` and we would need to work around it by doing:
```typescript
import fetchMock from 'fetch-mock';
type MockResponse = fetchMock.MockResponse;
```

It's not entirely clear to me whether this was a desired API change (esp. since you don't use typescript :)). If it wasn't, I am happy to send in a PR to fix it.

Thanks so much!

wheresrhys:
I didn't actually change the types aside from one or two lines related to ESM vs commonjs. Someone raised a bug with that which I thought was fixed last week https://github.com/wheresrhys/fetch-mock/issues/726.

What toolchain are you using? (typescript version, esm or commonjs, node version, testing framework)

brucespang:
It seems like this was introduced by 9ffba4019b76cd9899420e3a4f8ac3e3b3d42237 (switching from `export =` to `export default`). To reproduce, you should be able to add `import { MockRequest } from '..';` to the top of fetch-mock-tests.ts. For me, it compiles without problem with tsc 5.5.4 on commit ae1573996256a4ec1db05aa30452ebe80fdd3b33. Once I switch to 9ffba4019b76cd9899420e3a4f8ac3e3b3d42237, I run into the error

```
$ tsc types/fetch-mock-tests.ts
types/fetch-mock-tests.ts:2:10 - error TS2614: Module '".."' has no exported member 'MockRequest'. Did you mean to use 'import MockRequest from ".."' instead?

2 import { MockRequest } from '..';
           ~~~~~~~~~~~


Found 1 error in types/fetch-mock-tests.ts:2
```

[...]

PR #767, opened 2024-08-01
fix: commonjs module import from typescript

brucespang:
After our discussion #760, I went to update our imports and ran into the following problem:

Because `index.d.ts` has a default export, it seems like typescript decides that `packages/fetch-mock/dist/commonjs.js` should have a default export, but rollup does not create one:
```javascript
var index = FetchMock.createInstance();

module.exports = index;
```

This means that when we call `fetchMock.mock`, typescript turns this into `fetchMock.default.mock` which is undefined. If you want to see this behavior in a little more detail, [here's a repo](https://github.com/brucespang/fetch-mock-760-example) I used to debug it.

Rollup has an option [`output.exports`](https://rollupjs.org/configuration-options/#output-exports) which seems to fix it.

### DIFF
--- a/shared-rollup.config.js
+++ b/shared-rollup.config.js
@@ -6,6 +6,7 @@ export default {
 		dir: './dist',
 		entryFileNames: 'commonjs.js',
 		format: 'commonjs',
+		exports: 'named',
 	},
 	plugins: [
 		nodeResolve({ preferBuiltins: false }),


wheresrhys:
I've just merged https://github.com/wheresrhys/fetch-mock/pull/770 to add testing for importing into various toolchains. I've been able to reproduce the problem you're experiencing. Currently the job fails on the main branch because the tests are written to expect the change you've made in your PR.

This will technically be a breaking change so I'll need to release as v11, and will incorporate your change and documentation to describe it in a separate PR.

Thanks for investigating

brucespang:
Ah very nice, thank you!

PR #814, opened 2024-08-26
fix: correct Headers type on MockResponseObject interface

anstapol:


wheresrhys:
I don't think this is correct - [as documented](https://www.wheresrhys.co.uk/fetch-mock/docs/fetch-mock/API/Mocking/Parameters/response#headers), fetch-mock does expect an object, not a `Headers` instance, to be passed when creating a response. It might work if passed a `Headers` instance, but this is probably a fluke of hte internal implementation, and not something I want to explicitly support.

Can you explain why you need to pass in a `Headers` instance here?

anstapol:
> I don't think this is correct - [as documented](https://www.wheresrhys.co.uk/fetch-mock/docs/fetch-mock/API/Mocking/Parameters/response#headers), fetch-mock does expect an object, not a `Headers` instance, to be passed when creating a response. It might work if passed a `Headers` instance, but this is probably a fluke of hte internal implementation, and not something I want to explicitly support.
> 
> Can you explain why you need to pass in a `Headers` instance here?

You’re right, I haven’t checked the documentation yet. I reviewed the source code when I ran into the issue and discovered the Headers instance during debugging. It’s possible there might be an error in the docs?

anstapol:
image

I tried accessing the `Authorization` property on the object but encountered an error stating that it doesn’t exist. When I logged the result of `lastOptions().headers`, I found that it returned an instance of `Headers`.

wheresrhys:
I have found a [mistake in the types](https://github.com/wheresrhys/fetch-mock/pull/815) which will be having some effect, though I don't think it explains why you're getting a Headers instance with no Authorization header present. I woudl expect you to get a flat object containing all the headers passed to fetch

wheresrhys:
That change has now been released. Please give it a go

anstapol:
> That change has now been released. Please give it a go

Thanks for that. I think I'm still getting Headers instance...
Screenshot 2024-08-30 at 06 55 03
Screenshot 2024-08-30 at 06 54 27

anstapol:
Sorry if I seem confused. I know you’ve just updated the types, so the runtime code is the same. I just wanted to show you that this issue occurs in the latest version of the package, which is why I thought it might be a good idea to have the PR with the correct type (at least in my case). I’m not sure if it could behave differently in another setup, but it should be the same, right?

anstapol:
> That change has now been released. Please give it a go

Yes, that was the real issue—great catch! I’m now receiving the HeadersInit as expected. Thanks for your help!

PR #847, opened 2024-10-22
fix: force bodies which are in the middle of being read to be abortable

wheresrhys:


asrashley:
I created a temporary work-around using patch-package to modify the built code, as I was also finding my vitest tests failing due to an `abort()` call after the request had completed. To fix all of the test errors I needed to check both the request and the response bodies.

```patch
diff --git a/node_modules/fetch-mock/dist/esm/Router.js b/node_modules/fetch-mock/dist/esm/Router.js
index 599e79f..ef9ed5d 100644
--- a/node_modules/fetch-mock/dist/esm/Router.js
+++ b/node_modules/fetch-mock/dist/esm/Router.js
@@ -84,10 +84,19 @@ export default class Router {
                     const error = new DOMException('The operation was aborted.', 'AbortError');
                     const requestBody = request?.body || options?.body;
                     if (requestBody instanceof ReadableStream) {
-                        requestBody.cancel(error);
+                        if (requestBody.locked) {
+                                                       requestBody.getReader().cancel(error);
+                                               } else {
+                                                       requestBody.cancel(error);
+                                               }
                     }
-                    if (callLog?.response?.body) {
-                        callLog.response.body.cancel(error);
+                    const responseBody = callLog?.response?.body;
+                    if (responseBody) {
+                        if (responseBody instanceof ReadableStream && responseBody.locked) {
+                                                       responseBody.getReader().cancel(error);
+                        } else {
+                            responseBody.cancel(error);
+                        }
                     }
                     reject(error);
                 };
```

wheresrhys:
Closing in favour of #918

PR #866, opened 2024-11-08
fix: allow matching relative URLs

birtles:


wheresrhys:
Thanks for this. I'll use your PR as a starting point as I think it's highlighted an aspect of behaviour that is under-tested, and I'd like to add some more

wheresrhys:
closed in favour of https://github.com/wheresrhys/fetch-mock/pull/867

birtles:
> Thanks for this. I'll use your PR as a starting point as I think it's highlighted an aspect of behaviour that is under-tested, and I'd like to add some more

Thanks! Your PR looks a lot more thorough!

PR #871, opened 2024-11-13
Remove unsupported is-subset-of package

ottenhoff:


dmurvihill:
Looks like `is-subset`'s [test suite](https://github.com/b12-archive/is-subset/blob/master/test.js) could be pasted into fetch-mock and converted from jshint to vitest with a straightforward vim macro.

PR #878, opened 2024-11-22
Fix "Edit this page" links

midrissi:
Already proposed a fix in [this commit](https://github.com/wheresrhys/fetch-mock/commit/649e7e54e8278aa1a142a90a1597827ed4b286ff) but didn't notice that the **docs** folder has a nested **docs** folder


wheresrhys:
Thank you

PR #880, opened 2024-11-26
fix: adding types to jest matchers

hanseltime:
# Summary

This adds types to describe the FetchMockAdapters and updates the jest Expect object so that typescript and intellisense can provide correct typings.  It also adds some typings to the extension.js code so that, if something changes, we'll see that the types fail to compile and we can make updates as needed.

## Update - Additional Commit

This commit keeps support for the "patched fetch" and also adds support for FetchMock like the documentation details.  It adds tests for both inputs and adds an error message if the input is neither FetchMock or a patched fetch instance.

### DIFF
--- a/packages/jest/src/types.ts
+++ b/packages/jest/src/types.ts
@@ -72,11 +72,18 @@ export type FetchMockMatchers = {
 
 // types for use doing some intermediate type checking in extensions to make sure things don't get out of sync
 /**
+ * This reflects the Object.assign that FetchMock does on the fetch function
+ */
+export type PatchedFetch = {
+	fetchMock: FetchMock;
+};
+
+/**
  * This type allows us to take the Matcher type and creat another one
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type RawMatcher<T extends (...args: any[]) => any> = (
-	input: { fetchMock: FetchMock },
+	input: PatchedFetch | FetchMock,
 	...args: Parameters<T>
 ) => ReturnType<T>;
 
--- a/packages/jest/src/types.ts
+++ b/packages/jest/src/types.ts
@@ -1,0 +1,91 @@
+import type { CallHistoryFilter, FetchMock, UserRouteConfig } from 'fetch-mock';
+import type { SyncExpectationResult } from 'expect';
+
+export type HumanVerbs =
+	| 'Got'
+	| 'Posted'
+	| 'Put'
+	| 'Deleted'
+	| 'FetchedHead'
+	| 'Patched'
+	| 'Fetched';
+
+/**
+ * Verify that a particular call for the HTTP method implied in the function name
+ * has occurred
+ */
+export type ToHaveFunc = (
+	filter: CallHistoryFilter,
+	options: UserRouteConfig,
+) => SyncExpectationResult;
+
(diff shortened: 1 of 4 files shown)


wheresrhys:
This is really great - thanks so much for the contribution 🙌 

It will also need an update to the docs to explain that `fetch` can also be used. Are you happy to do that?

packages/jest/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+	| 'Patched'
+	| 'Fetched';
+
+/**
+ * Verify that a particular call for the HTTP method implied in the function name
+ * has occurred
+ */
+export type ToHaveFunc = (
+	filter: CallHistoryFilter,
+	options: UserRouteConfig,
```
Yihao-G, line 19: Pretty sure `filter` and `options` should be optional here?

packages/jest/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+) => SyncExpectationResult;
+
+/**
+ * Verify that a particular Nth call for the HTTP method implied in the function name
+ * has occurred
+ */
+export type ToHaveNthFunc = (
+	n: number,
+	filter: CallHistoryFilter,
+	options: UserRouteConfig,
```
Yihao-G, line 29: Same here

packages/jest/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+) => SyncExpectationResult;
+
+/**
+ * Verify that a particular call for the HTTP method implied in the function name
+ * has been made N times
+ */
+export type ToHaveTimesFunc = (
+	times: number,
+	filter: CallHistoryFilter,
+	options: UserRouteConfig,
```
Yihao-G, line 39: Same here

Yihao-G (commented):
As documented [here](https://www.wheresrhys.co.uk/fetch-mock/docs/wrappers/jest#:~:text=filter%20and%20options%20are%20the%20same%20as%20those%20used%20by%20fetchMock.callHistory.calls().):
> filter and options are the same as those used by fetchMock.callHistory.calls().

And `calls()`'s two arguments are optional:

https://github.com/wheresrhys/fetch-mock/blob/b63230f5d4334f96fdefc94ed75fe4dd7677109f/packages/fetch-mock/src/CallHistory.ts#L79

hanseltime:
> As documented [here](https://www.wheresrhys.co.uk/fetch-mock/docs/wrappers/jest#:~:text=filter%20and%20options%20are%20the%20same%20as%20those%20used%20by%20fetchMock.callHistory.calls().):
> 
> > filter and options are the same as those used by fetchMock.callHistory.calls().
> 
> And `calls()`'s two arguments are optional:
> 
> https://github.com/wheresrhys/fetch-mock/blob/b63230f5d4334f96fdefc94ed75fe4dd7677109f/packages/fetch-mock/src/CallHistory.ts#L79

@Yihao-G - if you think there is an issue with an already merged PR, I would encourage you to open an issue or submit your own PR.  People normally don't look at their old merged PRs

Yihao-G:
I just wanted to double confirm though 😄 - I'm happy to make a PR if you guys think those two parameters should be optional

PR #883, opened 2024-11-28
fix: route may be undefined when no match

RoCat:
When we just mock one route, if we have more requests than the one mocked, they dont have `.route` in callHistory. that result in JS error `(destructured parameter).route is undefined`

```
    fetchMock.mockGlobal();
    fetchMock.put('/documents', 200, 'updateDocuments');
    // remove line below and then its not working anymore
    fetchMock.route('*', 200);

    await $('>>>.chip.contract').click();
    await fetchMock.callHistory.flush(true);

    expect(
      fetchMock.callHistory.called('updateDocuments')
    ).toBe(true);
```

I may miss something, but the code change seems to fix this issue

### DIFF
--- a/packages/fetch-mock/src/CallHistory.ts
+++ b/packages/fetch-mock/src/CallHistory.ts
@@ -84,11 +84,11 @@ class CallHistory {
 
 		if (isMatchedOrUnmatched(filter)) {
 			if (([true, 'matched'] as CallHistoryFilter[]).includes(filter)) {
-				calls = calls.filter(({ route }) => !route.config.isFallback);
+				calls = calls.filter((call) => !call.route?.config.isFallback);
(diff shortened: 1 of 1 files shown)


wheresrhys:
Try using `fetchMock.catch()` instead of `fetchMock.route('*', 200)` - that would be the advised approach

Can I also ask what error you get exactly when you don't include the catch all route? Is it thrown from inside the call to `flush()`? fetch-mock should end up here https://github.com/wheresrhys/fetch-mock/blob/main/packages/fetch-mock/src/Router.ts#L208, but as you're using flush instead of await perhaps that error isn't bubbling up, making it a bit confusing

wheresrhys:
closing due to inactivity

PR #894, opened 2025-01-09
Change `lastCall` return type to `CallLog | undefined`

luhn:
This PR changes the return type of `CallHistory.lastCall` from `CallLog | void` to `CallLog | undefined`.

This matches the return type of [Array.pop](https://github.com/microsoft/TypeScript/blob/ea93ee6db9441eb51a1d816e114b0468363dcd9c/lib/lib.es5.d.ts#L1236)


wheresrhys:
thank you. Will release very soon

PR #921, opened 2025-03-03
fix: allow matching body for delete requests

pbomb:
Request bodies were ignored for DELETE requests (along with GET and HEAD requests) in #749, but DELETE requests can specify a body and it works with the `fetch-mock` library. As per [this MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch), only GET and HEAD requests cannot specify a body.

A test was added to verify that this works and this was also verified in a consuming library.

### DIFF
--- a/packages/fetch-mock/src/Matchers.ts
+++ b/packages/fetch-mock/src/Matchers.ts
@@ -212,7 +212,7 @@ const getBodyMatcher: MatcherGenerator = (route) => {
 	}
 
 	return ({ options: { body, method = 'get' } }) => {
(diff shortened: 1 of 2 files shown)


pbomb:
@wheresrhys Hey there. This is a fix for a bug that our team ran into. I'm not sure what the process is for contributing, so let me know if there's anything else you'd like me to do.

wheresrhys:
Great - thanks for contributing this. I'll get it released today

pbomb:
@wheresrhys Thank you so much!

PR #938, opened 2025-06-12
doc: update quickstart example

trietlam:
When I follow the example, I got error `the property ... does not exist on type FetchMock`

image

I believe now we have to access those method under the `callHistory` property instead of directly undef fetchMock, so update the doc accordingly


wheresrhys:
Thank you for this - will deploy soon

PR #944, opened 2025-08-25
fix: use optional chaining operator to prevent errors

stefan-schubert-sbb:


stefan-schubert-sbb:
Optional chaining is necessary, if a call is not mocked and in my opinion this should be possible. In this case a search through the call history should not fail and log an error or throw an named error to identify not defining a mock.
Logging a warning is better than failing completely in the case of aborting a locked request or response.

stefan-schubert-sbb:
@wheresrhys I added a test, which does fail without chaining.

wheresrhys:
Great - thanks for supplying this. I'll get it released ASAP

PR #953, opened 2025-10-23
fix: clean abort event listener once it's called or response is returned

Hilzu:
The event listener added for abort events is never cleaned. This causes issues if the signal aborts after the request is already processed and also leaks memory.

This can happen if `fetch` is used with the signal from `AbortSignal.timeout()`. The signal is aborted after the timeout even if the fetch request was already completed.

This change cleans the event listener after it is called with the `once` parameter. The `abort` event is only created once so this is safe.

The event listener is also cleaned after the response is created as it's not relevant anymore using the `signal` parameter and an abort controller. Alternative way to do this would be to lift the `abort` function higher in scope and remove the listener with `callLog.signal.removeEventListener('abort', abort);`.

These changes should fix many of the errors that occur regarding locked streams.

### DIFF
--- a/packages/fetch-mock/src/Router.ts
+++ b/packages/fetch-mock/src/Router.ts
@@ -156,6 +156,7 @@ export default class Router {
 		// eslint-disable-next-line no-async-promise-executor
 		return new Promise(async (resolve, reject) => {
 			const { url, options, request, pendingPromises } = callLog;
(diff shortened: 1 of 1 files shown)


wheresrhys:
Thanks @Hilzu  - this change makes a lot of sense and looks like it will fix the most annoying long-running bug on the project 🙌 .

I'll get it released today

